PR #331, opened 2025-10-06
Add `Subsets` (previously `Superset`)

### LINKED ISSUE #295: Add `Subsets` (previously `Superset`)

JCZuurmond:
# Changes

Resolves #295
Stacked on #327
Requires #336

Add `Subsets` that:
- Tracks dataset splits
- Archives and unarchives splits

# Checklist

### DIFF
--- a/src/proteingym/base/dataset.py
+++ b/src/proteingym/base/dataset.py
@@ -565,7 +565,7 @@ class Dataset(BaseModel):
 class SubsetsArchiveLayout:
     """The layout of a subsets archive."""
 
-    SUFFIX = ".splits.pgdata"
+    SUFFIX = f".splits{DatasetArchiveLayout.SUFFIX}"
     """The suffix of a subsets archive."""
 
     DATASET_ARCHIVE = f"dataset{DatasetArchiveLayout.SUFFIX}"


src/proteingym/base/dataset.py
```diff
@@ -52,9 +55,69 @@ class DatasetSlice:
...
+            slc: list[bool] | tuple[int | None, int | None, int | None],
+        ) -> list[bool] | slice:
+            """Convert an iterable to a slice object."""
+            if is_slice_iterable(slc):
+                return slice(*slc)
+            return slc
+
+        def as_dataset_slice(raw: dict) -> DatasetSlice:
+            assay_slices = [as_slice(slc) for slc in raw.get("assays", [])]
+            return cls(assays=assay_slices)
```
srivastavaprashant: I think that this could be one liner as well.

src/proteingym/base/dataset.py
```diff
@@ -52,9 +55,69 @@ class DatasetSlice:
...
+        ) -> list[bool] | tuple[int | None, int | None, int | None]:
+            """Encode a slice object as a tuple."""
+            if isinstance(slc, slice):
+                return (slc.start, slc.stop, slc.step)
+            return slc
+
+        def encode(dataset_slice: DatasetSlice) -> dict:
+            """Encode a DatasetSlice as a dictionary."""
+            slices = [encode_slice(s) for s in dataset_slice.assays]
+            return {"assays": slices}
```
srivastavaprashant: This could be done in fewer code lines as well

src/proteingym/base/superset.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Returns:
+            Path: The path to the dumped dataset archive.
+        """
+        if isinstance(path, str):  # User-friendly interface to support str
+            path = Path(path)
+        path = path or Path.cwd()
+        if path.is_dir():
+            path = path / f"{self.dataset.name}{SupersetArchiveLayout.SUFFIX}"
+        with ZipFile(path, "w") as zip:
```
srivastavaprashant: We already have more than one instance where we use a context. Does it make sense to create a separate context manager and use that avoiding duplication?

src/proteingym/base/superset.py
```diff
@@ -0,0 +1,105 @@
...
+            # While a SE practice is to avoid IO to disk where possible,
+            # we use a temporary directory here as long as dataset dump requires
+            # it.
+            with TemporaryDirectory() as temp_dir:
+                dataset_archive = self.dataset.dump(path=Path(temp_dir))
+                zip.write(
+                    dataset_archive, arcname=SupersetArchiveLayout.DATASET_ARCHIVE
+                )
+            slices_str = json.dumps([slc.to_json() for slc in self.slices])
```
srivastavaprashant: double encoding? one from `slc.to_json()` and one again with `json.dumps()`?

tests/conftest.py
```diff
@@ -0,0 +1,322 @@
...
+    sequence1 = Sequence(
+        name="seq1",
+        value=Seq("ACDEFG"),
+        type=SequenceType.WILD_TYPE,
+        alphabet=SequenceAlphabet.AA,
+    )
+    sequence2 = Sequence(
+        name="seq2",
+        value=Seq("GFEDCA"),
+        type=sequence1.alphabet,
```
srivastavaprashant:
```suggestion
        type=SequenceType.WILD_TYPE,
```

tests/test_superset.py
```diff
@@ -0,0 +1,98 @@
...
+    for _ in superset:
+        raise AssertionError("Should not iterate over empty superset")
+
+
+@pytest.mark.parametrize(
+    "slices, length",
+    [([], 0), ([slice(0, 1)], 1), ([True, False], 2)],
+)
+def test_superset_length(empty_dataset: Dataset, slices: list, length: int) -> None:
+    """Test that iterating over an empty superset does not yield any elements."""
```
srivastavaprashant: confusing docstring. Talks about element yield but the check is on length. 🤔

src/proteingym/base/superset.py
JCZuurmond: yeah, maybe. I am oke for now with having this context manager twice. Feels slightly premature to abstract this away. 

Also, I rather rework on avoiding the temporary directory context manager at all as the comment suggests that we want to avoid IO to disk

tests/test_superset.py
JCZuurmond: thanks, copied the test above, forgot to update the docstring

JCZuurmond (commented):
@srivastavaprashant : Could you please re-review? The PR became simpler with removing the `slice` support from `DatasetSlice`. I learned from the splitter implementation that slice will not commonly be used by splitters, masks are more natural for the use case. If we need the slice support in the future it can be added again

src/proteingym/base/superset.py
JCZuurmond: Yeah, I'm a slightly indifferent about this, but leaning to the current implementation. 

The above still holds: we implement (de)serialization methods on the classes, a dict is not guaranteed to be JSON serializable, hence we call `to_json` from the outside. 

The implementation on the inside (`DatasetSlice.to_json`) "knows" that for this specific class the JSON serialization is simply json dumping the dataclass converted to a dict.

I am indifferent as I am generally pro postponing decisions, and we did implement serialization already (`to_...` and `form_...`), but the `to_json` and `from_json` is new.

src/proteingym/base/superset.py
```diff
@@ -0,0 +1,99 @@
...
+
+    DATASET_ARCHIVE = f"dataset{DatasetArchiveLayout.SUFFIX}"
+    """The directory containing the dataset."""
+
+    SLICES_FILE = "slices.json"
+    """The file containing the slices."""
+
+
+@dataclasses.dataclass(kw_only=True, frozen=True)
+class Superset:
```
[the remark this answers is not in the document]
JCZuurmond: I created an issue for that: https://github.com/ProteinGym/proteingym-base/issues/350. Let's discuss today in-person.

JCZuurmond: @hredestig and @karel-w : I have been thinking about the name of this class. I think `Subset(s)`* would more accurate describe this implementation, see [wikipedia](https://en.wikipedia.org/wiki/Subset).

At the start, I was thinking about a more generic implementation of a "set of datasets" , hence `Superset`, but with the ask of [split strategies](https://github.com/ProteinGym/proteingym-base/issues/350) and the insight that a "set of datasets" can be just another dataset (e.g. one that combines many assays), this generic implementation made less sense.

I got fond of the name "Superset": who does *not* want "super" ⏫  in it's class? Lots more fun/high/exciting than "sub" ⏬ ... Still, the implementation became a(n iteration over) subset(s*).

*I am also reconsidering the plural v.s. singular form. With subset leaning towards the plural form opposed to the singular for "Superset" as there the "super" already suggest there is *more* than just the (data)set. Now with "Subset" it is not evident that the implementation is a container of multiple sets (dataset slices to be precise), it suggests there is *less* than a (data)set

PR #353, opened 2025-10-23
Add kfolds splitter

### LINKED ISSUE #352: Implement `KFoldsSplitter`
### Summary

Implement a `KfoldsSplitter`

### Motivation

Another common way to split datasets

### Proposed solution

_No response_

JCZuurmond:
# Changes

Resolves #352
Stacked on #342

Adds a KFoldsSplitter simlarly to https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html

# Checklist

### DIFF
--- a/tests/test_assay_operators.py
+++ b/tests/test_assay_operators.py
@@ -5,7 +5,6 @@ Module for testing assay operators.
 import pytest
 
 from proteingym.base.assay import Assay, AssayTarget
-from proteingym.base.assay import Assay, AssayTarget
 from proteingym.base.sequence import Sequence, SequenceAlphabet, SequenceType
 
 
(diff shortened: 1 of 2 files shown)


src/proteingym/base/splits.py
```diff
@@ -103,3 +103,55 @@ def split(self) -> Superset:
...
+        self.n_splits = n_splits
+        self.shuffle = shuffle
+
+    def split(self) -> Superset:
+        """Splits the dataset into K folds.
+
+        Returns:
+            Superset: The superset containing the splits.
+        """
+        records_shape = tuple(len(assay) for assay in self.dataset.assays)
```
JCZuurmond: @karel-w and @hredestig : The pattern that is chosen (as explained here for random splitter https://github.com/ProteinGym/proteingym-base/pull/342/files#r2455563247) is to manipulate the indices of the assay records to construct the splits. Let's discuss if this works for you.

We might be able to extract certain logic into a parent class and let the strategies focus on manipulating the indices only. But IMO it is slightly premature to do that now

karel-w:
Hmm one thing i'm thinking of for the benchmarking, do we now need to consider the split type within the model code now? Because that would be a bit inconvenient.

E.g. if we now have the train entrypoint, the model would need to know the amount of splits and how it is split?

```
def model(dataset: Superset):
    
    train_ds, val_ds, test_ds = tuple(dataset)

def train(dataset: Superset, model_card: ModelCard) -> 'MyModel':
    model = MyModel(model_card=model_card)

    model.train(dataset)
```

PR #370, opened 2025-11-13
Slice with target

### LINKED ISSUE #359: Add target to `DatasetSlice`
## Description

Add target to `DatasetSlice`

## Motivation

To slice columns next to records

## References

> this also need to pay attention to which target variables are in scope - otherwise you may end up with slices where you have only records for one target, and another slice for another target 🙃
> I guess something like, for every target in scope, first slice the records with sequences with values for all targets, then those with only n-1 targets, and so on.. 

 _Originally posted by @hredestig in [#342](https://github.com/ProteinGym/proteingym-base/pull/342/files#r2473142262)_

JCZuurmond:
# Changes

Resolves #359

Slice dataset with targets:
- Slice an assay with target(s)
- Slice dataset with targets:
  - Slices assays
  - Slices target names

# Checklist

### DIFF
--- a/src/proteingym/base/splits.py
+++ b/src/proteingym/base/splits.py
@@ -163,7 +163,7 @@ class RandomSplitter:
 
             assay_slices = []
             for mask, assay in zip(masks, dataset.assays, strict=True):
-                if targets:
+                if targets is not None:
                     if not any(target in assay.columns for target in targets):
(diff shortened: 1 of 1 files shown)


hredestig (approved):
Nice and it was educational to read. Don't have any suggestions for improvement!

PR #381, opened 2025-11-18
fix: dont shadow builtins

hredestig:
Here's an opinionated one: enable https://docs.astral.sh/ruff/rules/#flake8-builtins-a

Best practice is to not use names for variables that shadow python builtins. Let's make it a policy.

What do you think? 

# Changes

- format -> fmt
- zip -> zip_

# Checklist


JCZuurmond (dismissed):
Good! We could use a CI-check for this to avoid this happening in the futrue

JCZuurmond (approved):

PR #384, opened 2025-11-18
fixes bug382 by taking first for str targets

### LINKED ISSUE #382: .to_df() doesn't parse string-based targets
### Current behaviour

I've got a manifest as follows:

```
[[ assay_targets ]]
name = "activity"
description = "activity as measured by ..."

[[ assay_targets ]]
name = "residual_activity"
description = "residual activity measured ..."

[[ assay_targets ]]
name = "expression"
description = "expression of enzyme measured by ..."

[[ assay_targets ]]
name = "batchID"
description = "distinct batches"

[[ assays ]]
sequence = "sequence"
path = "cleaned_juniper_data.csv"
sequence_alphabet = "AA"
name = "all_targets"
[assays.targets]
"activity" = "activity"
"expression" = "expression"
"residual_activity" = "residual activity"
"batchID" = "batchID"
```
When calling dataset.to_df(), using e.g. batch ID: 'plate1-experimentA', batchID shows up as null values with column type str. When i converted batchIDs to numerical values at the CSV level and did not change the manifest, it parsed correctly.

The batch strings are correctly encoded at the assay level, e.g. with mock data:
```
[Assay(
 	name='all_targets',
 	description: None,
 	variables: 0,
 	records:
 		ACGTHYRMIHWMMEIDCNGCANNT, (0.10977630503022857, 171.6403790593661, 0.07210002816335391, 'batch_B'),
 		ACGTLSRRWNYDFWHKHVEQVKCY, (0.3742142217622615, 180.5291061253481, 0.00819118973473425, 'batch_B'),
 		ACGTRHNIRDQTEVIECGLEVNCL, (0.3350063246135029, 268.4720488212052, 0.21427566498924658, 'batch_C'),
 		...
 )]
```

### Desired behaviour

Being able to parse strings for categorical targets in to_df() functions

### Steps to reproduce

1. Create mock data resembling original data (Switch batch type from str/int by commenting correct row):
```
import pandas as pd
import numpy as np

np.random.seed(42)
n_rows = 100

df = pd.DataFrame({
    'sequence': [f'ACGT{"".join(np.random.choice(list("ACDEFGHIKLMNPQRSTVWY"), 20))}' for _ in range(n_rows)],
    'activity': np.random.normal(0.5, 0.2, n_rows),
    'expression': np.random.normal(200, 50, n_rows),
    'residual_activity': np.random.normal(0.15, 0.08, n_rows),
    # 'batch': np.random.choice(['batch_A', 'batch_B', 'batch_C'], n_rows)
    'batch' : np.random.choice([0, 1, 2], n_rows)
})

df.to_csv('mock_data.csv')
```

2. Create manifest
```
mock = """name = "Mock"
description = "Juniper dataset cleaned for platinum / MDGA citrate performance"
version = "1.0.0"

[[ assay_targets ]]
name = "activity"
description = "activity as measured by ..."

[[ assay_targets ]]
name = "residual_activity"
description = "residual activity measured by measuring activity, then heating proteins to 55C and measuring again after cooling down"

[[ assay_targets ]]
name = "expression"
description = "expression of enzyme measured by ..."

[[ assay_targets ]]
name = "batch"
description = "distinct biological batches"

[[ assays ]]
sequence = "sequence"
path = "mock_data.csv"
sequence_alphabet = "AA"
name = "all_targets"
[assays.targets]
"activity" = "activity"
"expression" = "expression"
"residual_activity" = "residual_activity"
"batch" = "batch"
"""

with open('mock.toml', 'w') as file:
    file.write(mock)
```

3. Load in dataset and return to df
```
manifest = Manifest.from_path('mock.toml')
dataset = Dataset.from_manifest(manifest=manifest)
dataset.to_df()
```

### Proposed solution

_No response_

karel-w:
# Changes

Resolves #382 

.to_df() resolves all targets as mean, even if the targets are str, leading to a column of null values. Since we accept str for AssayTarget we should return this in .to_df for e.g. non-numerical categorical targets. 

I've considered:
- not allowing str targets. Users would have to convert to e.g. numerical categories. Might lead to user A encoding e.g. 'EC:1.1.1.1' as 0, and user B encoding 'EC:1.1.1.100' as zero, might lead to confusion when the categorical targets are biological IDs.

- We convert it to numerical categorical targets. Might lead to confusion on iterating over the datasets, e.g. when reporting downstream in the benchmark could aggregate same numerical targets with different biological IDs. 

- The model converts str based categorical targets itself and we provide the biological ID. **(Current fix)**

# Checklist

### DIFF
--- a/tests/test_assay.py
+++ b/tests/test_assay.py
@@ -3,7 +3,6 @@ from string import ascii_uppercase
 from zipfile import ZipFile
 
 import polars as pl
-import polars.testing
 import pytest
 from Bio.Seq import Seq
 from pydantic import ValidationError


src/proteingym/base/dataset.py
```diff
@@ -588,13 +588,20 @@ def to_df(
             ~pl.all_horizontal([pl.col(t).is_null() for t in target_names])
         )
 
-        # Group by sequence and variables, and aggregate the target by mean
+        # Group by sequence and variables, and aggregate the target by mean,
```
hredestig: as discussed - another option is to have `aggfunc` and argument, which maybe a function that does you list below but also enables one to do something else

Also good log a warning if particularly the `.first` strategy is ever used since it lossy and probably most of the time a workaround rather than a solution.

tests/test_assay.py
```diff
@@ -535,20 +536,25 @@ def test_dataset_to_df_single_target(assay1: Assay, assay2: Assay) -> None:
...
-            df, expected_df, check_dtypes=False, check_column_order=False
-        )
+    # Aggregation function now returns warnings for hidden
+    # (default) behavior. Catching the warnings here as
+    # warnings get tested in test_dataframe_aggregation.py
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", UserWarning)
+        try:
+            df = dataset.to_df(target_names=["DMS Score"])
+        except ValueError as e:
```
hredestig: unnecessary try/except

tests/test_assay.py
```diff
@@ -561,22 +567,28 @@ def test_dataset_to_df_no_target(assay1: Assay, assay2: Assay) -> None:
...
-            }
-        )
-        pl.testing.assert_frame_equal(
-            df, expected_df, check_dtypes=False, check_column_order=False
-        )
+
+    # Aggregation function now returns warnings for hidden
+    # (default) behavior. Catching the warnings here as
+    # warnings get tested in test_dataframe_aggregation.py
+    with warnings.catch_warnings():
```
hredestig: dito / quiet warnings or unnecessary try/except

hredestig (dismissed):
Just the final warning filters, then great!

PR #391, opened 2025-11-27
Extend manifest with AssaysRaw

### LINKED ISSUE #170: Allow for both statistics and measurements in assays
Thanks for bearing with my scope creep here but,  I'm now we should track a file with statistics which has a list of annotated columns including the sequence (which  afaic is always present and always name "sequence") like this. I think we should additionally allow for a file with measurements, the data that was used to compute the statistics in a different file. Again with a sequence column and an arbitrary list of annotated columns.

```toml

[[ conditions ]]
name = "pH"
unit = "pH"
description = "value of pH"

[[ assays ]]
name = "activity"  # required
path = "statistics_assay_one.csv" # required

path_measurements = "measurements_assay_one.csv"  # optional

[ assays.conditions ]
condition_1 = <value>
condition_2 = <value>

[[ assays.statistics ]]
name = "performance over wild-type" # required
unit = ""        # optional
description = "pi" # optional

relative_to = "AQSVP.."    # optional, the sequence id of the molecule that the statistic is relative to. If present, it should with one of the sequences.
is_primary_target = true  # required, but only one of statistics can be true, thus default to `false`

[[ assays.statistics ]]
name = "area under the curve" 
description = "auc"     # enum of available possible statistics
is_primary_target = false

[[assays.measurements ]]
name = "OD"       # required
unit = nan            # optional 
description = ""  # optional

[[assays.measurements ]]
name = "concentration"
unit = "ppm"
description = ""
```

_Originally posted by @hredestig in https://github.com/ProteinGym2/pg2-dataset/pull/153#discussion_r2215633767_

JCZuurmond:
# Changes

Resolves #170 
Supersedes #366

Extend the manifest with a section for measurements, then handle it accordingly:
- Introduce `AssayRawManifestSection` to load the manifest section
- Introduce `AssayRaw` to hold the measurements data
- Introduce `Field` to track field/column metadata

Then update the `Dataset` and accompanying classes to include the raw assay data.

# Checklist

### DIFF
--- a/notebooks/02_Creating_Manifests_and_Archives.ipynb
+++ b/notebooks/02_Creating_Manifests_and_Archives.ipynb
@@ -402,7 +402,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -430,6 +430,7 @@
     "description = \"Binned DMS score\"\n",
     "\n",
     "[[ assays ]]\n",
+    "name = \"NEIME_2019\"\n",
     "sequence = \"mutated_sequence\"\n",
     "path = \"./NEIME_2019/Assays/Assay1.csv\"\n",
     "sequence_alphabet = \"AA\"\n",
@@ -854,12 +855,13 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 11,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
     "assays = [\n",
     "    AssayManifestSection(\n",
+    "        name=\"NEIME_2019\",\n",
     "        sequence=\"mutated_sequence\",\n",
     "        targets={\"DMS Score\": \"DMS_score\", \"DMS Score Bin\": \"DMS_score_bin\"},\n",
     "        sequence_alphabet=\"AA\",\n",


JCZuurmond (commented):
@hredestig : The plan is to first introduce the measurements related code. Then consolidate duplicate classes like the `Field` and `Measurement` <> `Assay`

docs/manifest.md
```diff
@@ -7,7 +7,7 @@ The manifest defines how a [dataset](data_model.md) is constructed.
 The following capabilities have been identified:
 
 | Capability           | Kind           | Motivation                                                |
-|----------------------|----------------|-----------------------------------------------------------|
+| -------------------- | -------------- | --------------------------------------------------------- |
```
hredestig: why this - here I see line goes to the end: https://www.markdownlang.com/extended/tables.html

I just approved and merged your MR to undo my earlier fix for this, read the change the wrong way 🙈

Anyway, pycharm complains and auto fixes to 
```
|----|---|
```
which is why i complained at this in the first place - you ide likes it like this?

src/proteingym/base/measurement.py
```diff
@@ -0,0 +1,194 @@
...
+
+class MeasurementsFormat(StrEnum):
+    """Supported assay file formats."""
+
+    CSV = ".csv"
+    """A comma separated text file"""
+
+
+@dataclasses.dataclass(kw_only=True, frozen=True)
+class Field:
```
hredestig: I find `Field` unfortunate in a package that uses pydantic. I proposed `AssayField` in #374 - can we merge that first? I know it may read double in the toml `assay.assay_field` instead of `assay.field` but stil better pydantic clash imo. 

- Column: no - refers to a 2D layout, says nothing.
- Variable: already use that
- Observable: bit cryptic
- Target: not necessarily a prediction target anymore
- AssayField: bit double since used with Assay
- DataField: ?
- ..

@karel-w

src/proteingym/base/measurement.py
```diff
@@ -0,0 +1,194 @@
...
+        columns = [field.name for field in section.fields]
+        # Reusing polars as we already depend on it for assays
+        records = list(pl.read_csv(section.path, columns=columns).iter_rows())
+        return cls(
+            name=section.name,
+            records=records,
+            fields=section.fields,
+            description=section.description,
+        )
+
```
hredestig: afford a `to_df` here as well - same thing as the one from assay but this time it's just a row-concatenation - and i guess a adding a column `name` to reflect the AssayRaw it came from

hredestig (commented):
Nice - mainly please revert to `AssayRaw` instead as discussed.

docs/manifest.md
JCZuurmond: Ah interesting, my editor does this when requesting a format. I do not care really as long as we are consistent

src/proteingym/base/assay.py
```diff
@@ -29,9 +30,57 @@ class AssayFormat(StrEnum):
...
+class Field:
+    """A raw assay field in an assay.
+
+    A field contains the metadata about a raw assay data, like the schema
+    definition of a dataset.
+
+    TODO
+    ----
+    Reuse this class across the code base.
+    Add field for setting the type.
```
JCZuurmond: Create issue for this

src/proteingym/base/dataset.py
```diff
@@ -141,6 +144,9 @@ class Dataset(BaseModel):
     assays: list[Assay] = Field(default_factory=list)
     """The assays present in the dataset."""
 
+    assays_raw: list[AssayRaw] = Field(default_factory=list)
```
JCZuurmond: Add test for dataset with assays raw

PR #396, opened 2025-12-01
feat: indicate how to best encode assay variables

### LINKED ISSUE #307: Assay conditions (variable) should indicate how to best be represented
### Summary

The data type can be int, str, float - but that doesn't give a clear indication how it should be represented - e.g. 1-hot encoded or direct regressor

### Motivation

Without this a variable like engineering_round could become an direct regressor - but it is clearly better thought of a categorical

### Proposed solution

add member `representation`(?) "categorical" or "numerical"

hredestig:
# Add annotation on assay variable for how to best encode it

Resolves #307 

If an assay variable is used as regression input, we need to know how to best encode it. I can think of two ways, as a categorical or a numerical but perhaps we need more distinctions so suggesting an enum here.

# Checklist

- [ ] I performed a self-review of my code. Amongst other things, I have commented my code in hard-to-understand areas.

### DIFF
--- a/src/proteingym/base/transformers.py
+++ b/src/proteingym/base/transformers.py
@@ -13,6 +13,7 @@ from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.compose import ColumnTransformer
 from sklearn.preprocessing import OneHotEncoder, StandardScaler
 
+from .assay import FieldEncoding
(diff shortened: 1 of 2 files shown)


hredestig:
One thing I was doubting though was whether the default should be numerical? Or better no default at all?

PR #405, opened 2025-12-03
feat: name a reference sequence

### LINKED ISSUE #23: Reference sequence
Description

Mark a specific sequence as reference sequence

Implementation

On dataset mention the reference sequence name, then also check if it exists

hredestig:
# Add a reference sequence to the top-level

Closes #23 
 
# Checklist

### DIFF
--- a/docs/decisions/0008-single-top-level-reference-sequence.md
+++ b/docs/decisions/0008-single-top-level-reference-sequence.md
@@ -1,0 +1,35 @@
+# 8. Single top-level reference sequence
+
+Date: 2025-12-04
+Status: Accepted
+
+## Context and Problem Statement
+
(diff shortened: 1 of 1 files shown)


src/proteingym/base/dataset.py
```diff
@@ -132,6 +132,12 @@ class Dataset(BaseModel):
     description: str | None = None
     """A brief description of the dataset."""
 
+    reference_sequence_name: str | None = None
```
karel-w: Maybe we could make a small ADR why we decided on only a single reference sequence at the top level and not multiple reference sequences.

karel-w (dismissed):

src/proteingym/base/dataset.py
```diff
@@ -132,6 +132,12 @@ class Dataset(BaseModel):
...
+    """Name of the sequence that is to be considered the reference for this dataset.
+
+    Useful for e.g. zero-shot models that compare likelihood for a token from a
+    reference with the token for a variant."""
```
JCZuurmond:
```suggestion
    reference with the token for a variant.
    """
```

JCZuurmond (dismissed):
Generally LGTM. I would appreciate a (small) PR on the top-level single reference sequence too.

PR #424, opened 2025-12-10
Predefined Splits

### LINKED ISSUE #421: Add ColumnSplitter
### Summary

Create a splitter that reads a column of split information, e.g. containing train/valid/validation/test strings or numbers representing grouped sets.

### Motivation

Required to fully port over PG1 to PG2

### Proposed solution

_No response_

karel-w:
# Changes

Resolves #421 

~~Ignore splits.py for now, want to check if required non-target makes sense this way.~~

# Checklist

### DIFF
--- a/src/proteingym/base/assay.py
+++ b/src/proteingym/base/assay.py
@@ -635,9 +635,7 @@ class Assay(AssayRaw):
         return cls(
             name=section.name or section.path.stem,
             records=records,
-            fields=[sequence_field]
-            + section.targets
-            + section.non_targets,
+            fields=[sequence_field] + section.targets + section.non_targets,
             description=section.description,
             variables=section.variables,
             non_targets=section.non_targets,


src/proteingym/base/assay.py
```diff
@@ -220,6 +220,9 @@ class AssayManifestSection(_ManifestSection):
     targets: list[Field] = pydantic.Field(default_factory=list)
     """The list of prediction targets in this assay."""
 
+    non_targets: list[Field] = pydantic.Field(default_factory=list)
+    """List of non-target fields that are included but not targets."""
```
hredestig:
```suggestion
    """List of non-target fields that are included but are not prediction targets."""
```

src/proteingym/base/assay.py
```diff
@@ -627,7 +640,12 @@ def as_manifest_section(self, *, path: Path) -> AssayManifestSection:
             description=self.description,
             sequence=self.fields[0].without_alias(),
             sequence_alphabet=sequence_alphabet,
-            targets=[f.without_alias() for f in self.fields[1:]],
+            targets=[
+                f.without_alias()
+                for f in self.fields
+                if f.name in self.target_feature_names
+            ],
+            non_targets=[Field(name=name) for name in self.non_target_feature_names],
```
hredestig: why not 
```
non_targets=self.non_targets
```

src/proteingym/base/splits.py
```diff
@@ -323,3 +323,88 @@ def split(self, dataset: Dataset, targets: list[str] | None = None) -> Subsets:
...
+        for split_value in split_values:
+            assay_slices = []
+            for assay in dataset.assays:
+                field_names = [f.name for f in assay.fields]
+                if self.split_column in field_names:
+                    col_idx = field_names.index(self.split_column)
+                    mask = [record[col_idx] == split_value for record in assay.records]
+
+                    if targets is not None:
+                        target_names = [e.name for e in assay.fields]
```
hredestig: redeclared (already computed with name `field_names` above)

src/proteingym/base/splits.py
```diff
@@ -323,3 +323,88 @@ def split(self, dataset: Dataset, targets: list[str] | None = None) -> Subsets:
...
+                    col_idx = field_names.index(self.split_column)
+                    all_values.update(record[col_idx] for record in assay.records)
+            split_values = self._sort_split_values(list(all_values))
+        else:
+            split_values = self.split_values
+
+        slices = []
+        for split_value in split_values:
+            assay_slices = []
+            for assay in dataset.assays:
```
hredestig: lots of indents here - bit tough to keep track of what's going on. Could help to early on
```
if self.split_column not in field_names:
    assay_slices.append(AssaySlice(records=[False] * len(assay), columns=[]))
    continue
```

karel-w: Added some comments, is this more clear?

docs/manifest.md
```diff
@@ -167,25 +170,32 @@ The assays section contains a list of assays included in the dataset.
...
 
 ``` toml
 [[assays]]
 path = "path/to/assay.csv"
-target = "DMS_score"
 sequence = "mutated_sequence"
+
+[[ assays.targets ]]
+name = "DMS Score"
+alias = "DMS_scorre
```
hredestig:
```suggestion
alias = "DMS_score"
```

src/proteingym/base/splits.py
```diff
@@ -323,3 +323,136 @@ def split(self, dataset: Dataset, targets: list[str] | None = None) -> Subsets:
...
+        if missing:
+            raise ValueError(
+                f"Dataset is missing required split values from split_order.\n"
+                f"  Missing keys: {sorted(missing)}\n"
+                f"  Observed keys: {sorted(observed)}"
+            )
+
+        return list(self.split_order)
+
+    def _validate_no_sequence_overlap(
```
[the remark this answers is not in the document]
karel-w: I think this should be in the scope of the ABC split class. I think we could wait till we have a few more split methods to determine the best patterns for the parent class. 

I've created an issue here to make sure we don't forget it: https://github.com/ProteinGym/proteingym-base/issues/429

PR #426, opened 2025-12-12
doc: ci-badge

hredestig:
badges!


karel-w (approved):
LGTM. Once its released and we got a release we could also consider adding a DOI (with badge) to each release: https://gist.github.com/seignovert/ae6771f400ca464d294261f42900823a

PR #427, opened 2025-12-15
refactor: provide only alias for sequence instead of whole field

### LINKED ISSUE #425: Sequence field in assay has become a bit funny
### Summary

`AssayManifestSection.sequence` became confusing.

### Motivation

We changed `assay.sequence` to a field instead of just a string referring to the column name in provided data file. That was nice because made targets align with fields. But as opposed to targets, where the richer `assay.targets` being fields, with types, unit etc, which are useful, the sequence does not benefit from being a field. More confusingly, the name of the sequence field must not actually be changed.

### Proposed solution

Make `AssayManifestSection.sequence_alias` a string, which results in `Assay.sequence == Field(name='sequence', alias=...)`

hredestig:
# Introduce sequence-alias instead of whole field in manifest

while we store the sequence as a field to make it symmetric with other field, it's not handy to enable user provide the whole field definition since type etc are not meaningful to change.

Resolves #425 

# Checklist

- [ ] I broke the PR down so that it contains a reasonable amount of changes for an effective review

### DIFF
--- a/docs/manifest.md
+++ b/docs/manifest.md
@@ -7,7 +7,7 @@ The manifest defines how a dataset is constructed.
 The following capabilities have been identified:
 
 | Capability           | Kind           | Motivation                                                |
-| -------------------- | -------------- | --------------------------------------------------------- |
+|----------------------|----------------|-----------------------------------------------------------|
 | Primitive data types | Functional     | For data representation, like `int`, `bool` and `string`. |
 | Lists                | Functional     | For organizing data entries.                              |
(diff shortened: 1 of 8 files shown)


karel-w (approved):
LGTM, very small nitpick on assay line 607.

Also came across that we don't explain each section fully in the ipynb, but i think its outside the scope of this PR so made an issues here: https://github.com/ProteinGym/proteingym-base/issues/428

karel-w (approved):

PR #460, opened 2026-05-06
swapping biopython MSA handling for evedesign.sequence.Sequences

ethane4:
# Changes

Resolves inability to handle MSAs w/ insertion states by relying on evedesign.sequence.Sequences instead of bio.Align module.

- I also changed the implicit MSA format assumptions to .a3m instead of .fasta
- Tried to update tests where applicable
- evedesign package should add minimal dependency overhead--in the newest release, we removed the torch dependency, so the base package w/ no extra options is super light 

# Checklist

- [ x] I broke the PR down so that it contains a reasonable amount of changes for an effective review
- [x ] I performed a self-review of my code. Amongst other things, I have commented my code in hard-to-understand areas.
- [ x] I made corresponding changes to the documentation
- [ x] I added tests that prove my fix is effective or that my feature works
- [ x] I accounted for dependent changes to be merged and published in downstream modules


src/proteingym/base/msa.py
```diff
@@ -24,6 +23,9 @@ class MSAFormat(StrEnum):
     FASTA = "fasta"
```
karel-w: Should we drop fasta and assume A3M always?

src/proteingym/base/msa.py
```diff
@@ -186,7 +188,7 @@ class MSA:
     name: str
     """The name of the MSA."""
 
-    value: MultipleSeqAlignment
+    value: Sequences
     """The value of the MSA, typically a file path or binary data."""
```
karel-w: Can this still be a file path or binary data?

karel-w (approved):
We could consider dropping fasta format for alignments.

Please update the documentation in the README and double check if we need to update the documentation in the notebooks. 

Also apply linting with `ruff check src/` and `ruff check tests/`.

Then the CI should pass, feel free to merge after.

src/proteingym/base/msa.py
ethane4: evedesign.sequence.Sequences can support a3m, a2m, fasta, and fasta_unaligned. For the PG2 database, everything will be a3m. I guess we can try to force people to always use a3m - I will just remove the fasta option.

src/proteingym/base/msa.py
ethane4: fixed.

karel-w (approved):

pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
     "scikit-learn>=1.5.0,<2.0.0",
     "pandas>=2.3.3",
     "requests>=2.32.3",
+    "evedesign>=0.0.4",
```
hredestig, line 23: This (eventually) causes a circular dependency since evedesign should (imo) be able to convert pg-base datasets to its preferred format. Also, evedesign is a heavy dependency which makes it harder to use for diverse environments. To me it would make most sense if proteingym-base was imported by evedesign for definitions of classes on datasets, design objectives and not the other way around.

src/proteingym/base/msa.py
```diff
@@ -4,8 +4,7 @@
 from typing import Any
 
 import numpy as np
-from Bio import AlignIO
-from Bio.Align import MultipleSeqAlignment
+from evedesign.sequence import Sequences
```
hredestig, line 7: Move this class to proteingym-base instead? @thomashopf ?

pyproject.toml
thomashopf: Yes, fully agree re this approach. We can add `System.from_pgdata()` and `LabeledInstanceDataset.from_pgdata()` methods to evedesign to transform the data as needed. Then could call these in pg-benchmark to instantiate the models.

Independently, working on making the core evedesign package a lighter dependency. The latest release does not depend on pytorch anymore unless you add a model option that needs pytorch, and considering the same for numba going forward.

src/proteingym/base/msa.py
thomashopf: My take for simplicity here would be to keep the MSA and 3D structures as files, and then let the consumer (e.g. `System.from_pgdata()`, cf. comment above) decide how to load them from the pgdata?

pyproject.toml
ethane4: I guess this would probably mean that the value attribute of the MSA class in pg-base would have to be None by default? Alternatively we could come up with some lightweight method for storing MSAs w/ insertion states (ex. loading in fasta-based alignment formats as dicts/arrays or something, and then handling the conversion to Sequences() downstream in evedesign)

src/proteingym/base/msa.py
hredestig: Only keeping files makes validation complicated though. I.e., I think it would be valuable that I could be sure that for a pgdataset, the reference sequence 'matches' the MSA(s) and the structure - but for that you have to load the files and inspect them. Perhaps we can do both - have a pgdata native class, and the source file still available.

thomashopf: I think that dual approach makes a lot of sense

pyproject.toml
hredestig: allowing value to be none seems a bit confusing to me - but perhaps we can consider another attribute here - `MSA.source_file: typing.BinaryIO` - and then let downstream packages that already have its own `from_xyz` methods simply do `from_xyz(data.msas[0].source_file)`

src/proteingym/base/msa.py
ethane4: Any preference on the pgdata native MSA handling? I guess we will drop insertion states to verify the query sequences in the MSAs match the structures, so we could just strip the insertions for fasta-based formats if we don't plan on using the pgdata native value datatypes for anything

Issue #448, opened 2026-03-31
Reference sequence relevant region


hredestig:
### Summary

Enable us to annotate a special region on a sequence to deal with the fact that not always the whole natural protein is what is being assayed, included in MSA and in the structure etc. 

### Motivation

A protein may e.g. a terminal pro region that is cleaved off during maturation. An evolutionary trained LLM may benefit from the entire context, but only the mature protein is subjected to experimental assays. 

Also MSAs may be done on only part of sequence e.g. a special domain of interest, and structure may not be available for certain regions. 

### Proposed solution

Add region field to `Sequence` and `SequenceManifestSection`
```python
class SequenceRegion(BaseModel):
    start: PositiveInt
    length: PositiveInt

class SequenceManifestSection(BaseModel):
    region: SequenceRegion | None
    ...
```
and associated properties for getting the sliced out relevant region.

Would we ever need more than one region?

hredestig:
What's a good example dataset that depends on this feature?

hredestig:
Also related - perhaps to open in different issue:

Thomas: MSA-based models typically expect their reference sequence to match the target sequence in the MSA; a similar requirement holds for 3D structure-based models. There are only 2 examples where the MSA region is different from the reference sequence in PG1, but this most likely will increase as running MSAs on full reference sequences is a point of criticism with PG1. One can easily slice the reference sequence and mutant sequences to the MSA subregion, but is that an operation that should be applied for all models, or just upon request for MSA-based models, e.g. with an extra param to train()? (Note there is a lot of substitution mapping going on internally in PG1 model wrappers for MSA-based models dealing exactly with this matter)

florisvdf:
> What's a good example dataset that depends on this feature?

Currently all datasets in PG1 have matching `target_seq` length and variant sequence length. But I think the Tsuboyama datasets are good examples of datasets where this feature would be useful, since their variant sequences and `target_seq` correspond to individual domains of proteins. For reasons you mentioned before, storing the reference sequence of the entire protein combined with a region that matches the variant sequences would be useful.

PR #465, opened 2026-05-12
adding descriptors to manifest fields

ethane4:
# Changes

Adding some additional descriptors to parts of the manifest (to align with the info we have been collecting for new pg2 datasets)

- overall has substitutions/indels/multiples attributes
- pfam and start index attributes for sequences
- mutated region start/end + assay categories to assay data 
- variant count is inferred from the assay csvs

# Checklist


src/proteingym/base/assay.py
```diff
@@ -25,6 +25,7 @@
 from .sequence import Sequence, SequenceAlphabet, SequenceType
 
 SEQUENCE = "sequence"
+MUTANT = "mutant"
```
florisvdf: I think it would make sense to always have a sequence column for variant sequences. If the variant sequences are present already, is there a reason to also include mutant information?

ethane4: I think the big thing would be counting the mutation orders of individual variants and being able to quickly understand what seqs have indels/frameshifts etc. Combined w/ the start index/mutation index manifest fields there should be no ambiguity about what indices the mutation labels refer to

example_data/neime_2019.toml
```diff
@@ -1,6 +1,9 @@
 name = "NEIME_2019"
 description = "The NEIME Kennouche 2019 (UniProt id: A0A1I9GEU1) datase"
 version = "1.0.0"
+has_substitutions = true
+has_indels = false
+has_multiples = false
```
florisvdf: Could we do the same as for variant count and automatically infer these from the assay csvs instead of needing to declare them?

src/proteingym/base/assay.py
florisvdf: Makes sense. I think another option could be to have this automatically assigned with a mapping function like `to_mutant`. Upside of that is that it reduces bloat of the `manifest.toml` and would not require any validation.

example_data/neime_2019.toml
```diff
@@ -29,6 +32,12 @@ name = "NEIME_2019"
 path = "./NEIME_2019/Assays/Assay1.csv"
 sequence_alphabet = "AA"
 sequence_alias = "mutated_sequence"
+mutated_region_start = 1
```
karel-w: Do we need to know where along the sequence the mutated region starts/ends? Wouldn't we assume all sequences are mutated besides our sequence we set as reference sequence?

example_data/neime_2019.toml
```diff
@@ -29,6 +32,12 @@ name = "NEIME_2019"
...
+mutated_region_end = 161
+readout = "sequencing"
+library_construction_method = "random mutagenesis"
```
karel-w: How do you want to validate the readout, library_construction_method, assay_method and transformation? Only allow for a set of options? Otherwise might be a case for an [[ assays.metadata ]] subsection where we can allow any key-value to be set?

example_data/neime_2019.toml
```diff
@@ -50,6 +59,8 @@ path = "./NEIME_2019/sequences/A0A1I9GEU1.fasta"
 type = "wild_type"
 alphabet = "AA"
 uniprot_id = "A0A1I9GEU1"
+pfams = "PF00114"
+start_index = 1
```
karel-w: What is the start index of an assay target?

example_data/neime_2019.toml
```diff
@@ -50,6 +59,8 @@ path = "./NEIME_2019/sequences/A0A1I9GEU1.fasta"
 type = "wild_type"
 alphabet = "AA"
 uniprot_id = "A0A1I9GEU1"
+pfams = "PF00114"
```
karel-w: pfam is not related to the assay target right?

src/proteingym/base/assay.py
karel-w: Wouldn't it be possible to calculate this as a property from the mutated sequence and the reference sequence and not have to set all this up front? Or does that not work for frameshifts?

src/proteingym/base/assay.py
```diff
@@ -235,9 +236,36 @@ class AssayManifestSection(_ManifestSection):
...
+    transformation: str | None = None
+    """The transformation method."""
+
+    assay_method: str | None = None
+    """The assay method."""
+
+    target_phenotype: str | None = None
+    """The target phenotype of the assay."""
+
+    variant_count: int | None = None
```
karel-w: Could be a property calculated from the sequences?

src/proteingym/base/assay.py
karel-w: Also we assume that all the sequences except those set in the Sequence object are already mutants

src/proteingym/base/assay.py
```diff
@@ -457,6 +489,33 @@ class Assay(AssayRaw):
...
+    """The number of variants in the assay."""
+
+    mutant_alias: str | None = None
```
karel-w: How is this different from sequence_alias?

karel-w (commented):
We need to discuss what we store and what can be stored as properties calculated from what is stored.

src/proteingym/base/assay.py
ethane4: I am working on writing out fxns to cover this. In our meeting on Friday we discussed limiting the user-defined dataset properties as much as possible and just inferring information from what is provided by the user (ex. some function that creates mutation strings given a mutated sequence and a parent sequence). The insertions/deletions will probably actually require some pairwise alignment, but substitutions will be easy to infer.

example_data/neime_2019.toml
ethane4: If we are inferring the mutant labels from the parent/mutated sequence we will need to know the uniprot index of the first amino acid in the parent sequence to construct accurate labels (ex. it would be preferable for everything in the dataset to correspond to the uniprot entry - if there is a uniprot entry)

example_data/neime_2019.toml
ethane4: In some cases, we have proteins that are ex. 1000 amino acids long with two regulatory domains and one catalytic domain and the experimentalists will perform three separate assays on each domain (like mutate one domain, splice it back into the full length protein, and then measure the overall protein function).

Ex. we would like to know the difference between two datasets that may otherwise appear identical at the manifest level.

example_data/neime_2019.toml
ethane4: We have not come up with the full control vocabulary for each of these options yet, but I think we have a pretty decent idea. I can add some validation checks at some point

src/proteingym/base/assay.py
ethane4: Variant count is calculated from the assay csv length (if that is what this is referring to) 

Target phenotype will have to be user specified ex. abundance, activity, etc.

src/proteingym/base/assay.py
ethane4: sequence_alias is the column of the assay csv that corresponds to the mutated seqs (I think... need to go back and check)

mutant_alias would be the column w/ the mutant strings

example_data/neime_2019.toml
ethane4: What do you mean?

example_data/neime_2019.toml
florisvdf: If we would like to use these descriptors to be able to group benchmarking results like in the original ProteinGym paper, putting them in an unconstrained metadata field will not be sufficient.

example_data/neime_2019.toml
florisvdf: Not sure what you mean by "assay target" here @karel-w . Also I wonder if we can reliably infer this. Doing would require alignment I guess which may not always yield a correct index, especially for short sequences. Otherwise this should be a required field?

ethane4: Yeah I was going to say we could infer if there is a uniprot_id present but it may add a lot of unnecessary overhead ex. pulling the uniprot entry from a database (or the API), pairwise alignment, etc. when the user could just provide it

src/proteingym/base/assay.py
ethane4: After thinking about this a bit more, it may be really annoying to write out a function that does this properly (probably an extra ~50-100 lines of code).

I think for downstream analysis in the PG2 paper, we will need to have the mutant information (to subset data by mutation order, presence of indels, etc.), so including it may be unavoidable. 

It would almost be preferable to take a dataset at construction time and add an independent Assay manifest section for insertions, deletions, and each substitution order ex. singles, doubles, ... This would allow us to avoid explicitly storing the mutation strings somewhere (since all we are really concerned with is the type of mutant). In the end, we will have to split the datasets like this when we are calculating metrics anyways.

example_data/neime_2019.toml
hredestig: I'm curious if this could not also better be expressed as a computed property, or a function. 

- Explicit value should better be validated - result of a function is necessarily valid (assuming correct implementation)
- Explicit value complicates future enhancements as possibility to add two datasets together
- Exact definition of `mutated_region_start` is not clear to me. I can imagine what several options, and read documentation to see what is intended, still it opens option for difference in opinion which also needs to be reconciled

src/proteingym/base/assay.py
hredestig: > Combined w/ the start index/mutation index manifest fields there should be no ambiguity about what indices the mutation labels refer to

Ambiguity can arise when you share or in any way separate the mutation string from the original context. I see that mutation strings are safe when they are representation of an object that contains the aligned parent sequence, and as such great to inspect diversity, but when merely string literals, they are a concern. 

I'm curious to hear more context of when you need mutation strings?

src/proteingym/base/assay.py
```diff
@@ -235,9 +236,36 @@ class AssayManifestSection(_ManifestSection):
...
     sequence_alphabet: SequenceAlphabet | None = None
     """The alphabet of the sequences of the assay."""
 
+    mutated_region_start: int | None = None
+    """The start index of the mutated region, 1-indexed."""
+
+    mutated_region_end: int | None = None
+    """The end index of the mutated region, 1-indexed."""
+
+    library_construction_method: str | None = None
```
hredestig: I think better if this was an enum - or you will have no control over spelling errors / synonyms

src/proteingym/base/assay.py
```diff
@@ -235,9 +236,36 @@ class AssayManifestSection(_ManifestSection):
...
+    """The method used for library construction."""
+
+    readout: str | None = None
```
hredestig: enum

src/proteingym/base/assay.py
```diff
@@ -235,9 +236,36 @@ class AssayManifestSection(_ManifestSection):
...
+    """The readout of the assay."""
+
+    transformation: str | None = None
```
hredestig: enum

src/proteingym/base/assay.py
```diff
@@ -235,9 +236,36 @@ class AssayManifestSection(_ManifestSection):
...
+    """The transformation method."""
+
+    assay_method: str | None = None
```
hredestig: enum - and more documentation of what is meant here please

src/proteingym/base/assay.py
```diff
@@ -235,9 +236,36 @@ class AssayManifestSection(_ManifestSection):
...
+    """The assay method."""
+
+    target_phenotype: str | None = None
```
hredestig: This appears redundant with the `targets` below?

src/proteingym/base/assay.py
hredestig: Is this `len(set(r[0] for r in assay.records)` ? Or as implied below just `len(assay.records)`? If the latter, the `number_of_records` may be better name and either way these could better be computed properties?

src/proteingym/base/assay.py
```diff
@@ -235,9 +236,36 @@ class AssayManifestSection(_ManifestSection):
...
     """The sequence feature name given in the file."""
 
+    mutant_alias: str | None = None
+    """The mutant feature name given in the file."""
+
     sequence_alphabet: SequenceAlphabet | None = None
     """The alphabet of the sequences of the assay."""
 
+    mutated_region_start: int | None = None
+    """The start index of the mutated region, 1-indexed."""
```
hredestig: 1-indexing is potential source of nasty bugs.. since we use python, could we just stick to 0-indexing? (if this property is absolutely needed, I'm still keen to see if we couldn't make all this computed)

hredestig (changes requested):

example_data/neime_2019.toml
karel-w: So just for understanding, we want to align our variant sequences to the start_index of the uniprot sequence to get mutation labels based on the uniprot id?

I agree alignments can change the automatic way of notating this, but it should be a property of the reference sequence and not the assay target then? Unless its different for different sequences in our assay, but then I would just capture it from a column. 

Also FYI, we already do pull data from the uniprot when constructing a dataset for the taxonomic in our Sequence class.

example_data/neime_2019.toml
karel-w: Shouldn't it be a property of the [Sequence](https://github.com/ProteinGym/proteingym-base/blob/060cf1b84ab131039cdb9a281b0e4d5370462240/src/proteingym/base/sequence.py#L161) and not the assay? Here we also mark the reference / wild type sequence.

example_data/neime_2019.toml
karel-w: > We have not come up with the full control vocabulary for each of these options yet, but I think we have a pretty decent idea. I can add some validation checks at some point

If we have a pretty decent idea It makes sense to add a vocabulary as enum such that we validate the groupings, otherwise currently we still have it similar to an unconstrainted metadata field, e.g. people can add "random mutagenesis" or "mutagenesis" refering to the same thing, but not being groupable. 

We have e.g. the same for [SequenceType](https://github.com/ProteinGym/proteingym-base/blob/060cf1b84ab131039cdb9a281b0e4d5370462240/src/proteingym/base/sequence.py#L33), where we allow only a few specific defined types, and can extend them if needed.

src/proteingym/base/assay.py
ethane4: I suppose we do not need the actual mutation string, but we will need to be able to subset datasets by mutation order/insertions/deletions for downstream analysis. Scoring indels w/ masked log likelihoods may also require some strange data manipulations that may be easier if we know where the indels are. 

The bottom line would just be what I was saying above w/ having some method to subset by mutation order, etc. If you guys can think of a smooth way to do this, let me know.

example_data/neime_2019.toml
florisvdf: Probably deserves it's own PR. See https://github.com/ProteinGym/proteingym-base/issues/448

example_data/neime_2019.toml
ethane4: This is resolved now I think?

src/proteingym/base/assay.py
```diff
@@ -46,6 +46,94 @@ class FieldEncoding(StrEnum):
     NUMERICAL = "numerical"
 
 
+class LibraryConstructionMethod(StrEnum):
+    """The method used to construct the protein library.
+
+    random mutagenesis: uncontrolled introduction of
```
hredestig: small nitpick here - in the other `StrEnum` (and all other classes as well) we write the docstrings like

```python
class MyEnum(StrEnum):
    """Title.

    Description
    """

    my_attr: int
    """Description of my attr"""
```

which makes it render nicely in the webpage

src/proteingym/base/assay.py
```diff
@@ -46,6 +46,94 @@ class FieldEncoding(StrEnum):
...
+
+    none: raw read counts, product amounts, etc.
+
+    non-parametric: some basic data transformation applied
+    (ex. ratio of read counts, log transformation, etc.)
+
+    fit: some regression/ML-based fit applied to data
+    (ex. MoCHI, Enrich2, etc.)"""
+
+    NONE = "None"
```
hredestig: nitpick - "none" (or is there a reason for the capital here)

src/proteingym/base/assay.py
```diff
@@ -46,6 +46,94 @@ class FieldEncoding(StrEnum):
...
+
+    activity: some chemical reaction the protein performs
+
+    abundance: measurement of protein/RNA abundance
+
+    expression: measurement of protein expression
+
+    stability: thermostability measurements
+
+    binding: protein binding to sometehing
```
hredestig: typo: something

src/proteingym/base/sequence.py
```diff
@@ -92,6 +92,9 @@ class SequenceManifestSection(BaseModel):
     uniprot_id: str | None = None
     """The UniProt identifier for this sequence."""
 
+    pfams: str | None = None
```
hredestig: strings? should it be a list then?

tests/test_assay.py
```diff
@@ -329,6 +333,45 @@ def test_assay_manifest_section(assay_file: Path) -> None:
         assert all(target.alias_ in content for target in section.targets)
 
 
+def test_assay_manifest_section_descriptors(assay_file: Path) -> None:
```
hredestig: this test looks like testing that python dataclasses work as intended and hence redundant

src/proteingym/base/assay.py
```diff
@@ -473,6 +591,11 @@ def target_feature_names(self) -> list[str]:
         non_target_names = [f.name for f in self.non_targets]
         return [name for name in all_names if name not in non_target_names]
 
+    @property
+    def number_of_variants(self) -> int:
+        """Returns the number of variants in the assay."""
+        return len(self)
```
hredestig: shouldn't this be named `number_of_records`? or should it be 
```
len({e[0] for e in self.records})
```
..to count the number of (unique) sequences?

tests/test_assay.py
```diff
@@ -329,6 +333,45 @@ def test_assay_manifest_section(assay_file: Path) -> None:
...
+    assert section.readout == AssayReadout.DNA_SEQUENCING
+    assert (
+        section.library_construction_method
+        == LibraryConstructionMethod.RANDOM_MUTAGENESIS
+    )
+    assert section.assay_method == AssayMethod.SELECTION
+    assert section.transformation == AssayTransformation.NONE
+
+
+def test_assay_from_manifest_section_with_descriptors(assay_file: Path):
```
hredestig: same as above

tests/test_assay.py
```diff
@@ -546,6 +589,31 @@ def test_assay() -> None:
...
+def test_assay_number_of_variants():
+    seq1 = Sequence(
+        name="seq1",
+        value=Seq("APC"),
+        type=SequenceType.STANDARD,
+        alphabet=SequenceAlphabet.DNA,
+    )
+    seq2 = Sequence(
+        name="seq2",
+        value=Seq("DEF"),
```
hredestig, line 573: test would pass even if the sequence was "APC" here as well right?

hredestig (changes requested):
Nice! Just a few minor comments

src/proteingym/base/assay.py
ethane4: I guess in some assays # of unique sequences != # of measurements ex. sometimes the same seq is measured 10 times. I guess we are more concerned with # of unique seqs?

tests/test_assay.py
ethane4: went w/ your suggestion above to avoid this issue.

tests/test_assay.py
ethane4: removed

tests/test_assay.py
ethane4: removed

src/proteingym/base/sequence.py
ethane4: swapped to list

src/proteingym/base/assay.py
ethane4: fixed

src/proteingym/base/assay.py
ethane4: made lowercase

src/proteingym/base/assay.py
ethane4: fixed all of them

src/proteingym/base/sequence.py
```diff
@@ -180,6 +183,9 @@ class Sequence:
     uniprot_id: str | None = None
     """The UniProt identifier for this sequence."""
 
+    pfams: list | None = None
```
hredestig: we call it `uniprot_id` so analogously maybe nicer with `pfam_ids`?
```suggestion
    pfam_ids: list[str] | None = None
```

src/proteingym/base/assay.py
```diff
@@ -29,21 +29,129 @@
...
+
+    RNA_SEQUENCING = "rna_sequencing"
+    """rna_sequencing: fitness is estimated from RNA sequencing"""
+
+    DNA_SEQUENCING = "dna_sequencing"
+    """dna_sequencing: fitness is estimated from DNA sequencing"""
+
+    FLUORESCENCE = "fluorescence"
+    """fluorescence: fitness is estimated from product and/or
+    reporter fluorescence"""
```
hredestig:
```suggestion
    """Fluorescence
    
    Fluorescence fitness is estimated from product and/or
    reporter fluorescence"""
```

src/proteingym/base/assay.py
```diff
@@ -29,21 +29,129 @@
...
+
+class LibraryConstructionMethod(StrEnum):
+    """LibraryConstructionMethod
+
+    The method used to construct the protein library.
+    """
+
+    RANDOM_MUTAGENESIS = "random_mutagenesis"
+    """random mutagenesis: uncontrolled introduction of
+    mutations across random positions (ex. epPCR)"""
```
hredestig:
```suggestion
    """Random mutagenesis
    
    Uncontrolled introduction of mutations across random 
    positions (ex. error-prone PCR)."""
```

src/proteingym/base/assay.py
```diff
@@ -29,21 +29,129 @@
...
+
+    SITE_SATURATED_MUTAGENESIS = "site_saturated_mutagenesis"
+    """site-saturated mutageneisis: introduction of mutants
+    at specific sites with complete coverage (ex. NNK/S)"""
```
hredestig:
```suggestion
    """Site-saturated mutagenesis
    
    Introduction of mutants at specific sites with complete coverage (ex. NNK/S)"""
```

src/proteingym/base/assay.py
```diff
@@ -29,21 +29,129 @@
...
+
+    DISCRETE = "discrete"
+    """discrete: designed libraries targeting specific positions
+    and specific amino acid coverage (ex. oligo pools)"""
```
hredestig:
```suggestion
    """Discrete
    
    Designed libraries targeting specific positions and specific 
    amino acid coverage (ex. oligo pools)"""
```

src/proteingym/base/assay.py
```diff
@@ -29,21 +29,129 @@
...
+    DNA_SEQUENCING = "dna_sequencing"
+    """dna_sequencing: fitness is estimated from DNA sequencing"""
+
+    FLUORESCENCE = "fluorescence"
+    """fluorescence: fitness is estimated from product and/or
+    reporter fluorescence"""
+
+    PRODUCT = "product"
+    """product: fitness is estimated from product concentration
+    measured in some manner other than fluorescence"""
```
hredestig:
```suggestion
    """Product 
    
    Fitness is estimated from product concentration
    measured in some manner other than fluorescence."""
```

src/proteingym/base/assay.py
```diff
@@ -29,21 +29,129 @@
...
+
+    The transformation applied to the raw assay data.
+    """
+
+    NONE = "none"
+    """none: raw read counts, product amounts, etc."""
+
+    NON_PARAMETRIC = "non_parametric"
+    """non-parametric: some basic data transformation applied
+    (ex. ratio of read counts, log transformation, etc.)"""
```
hredestig:
```suggestion
    """Non-parametric
    
    Some basic data transformation applied
    (ex. ratio of read counts, log transformation, etc.)"""
```

src/proteingym/base/assay.py
```diff
@@ -29,21 +29,129 @@
...
+
+    FIT = "fit"
+    """fit: some regression/ML-based fit applied to data
+    (ex. MoCHI, Enrich2, etc.)"""
```
hredestig:
```suggestion
    """Fit
    
    Some regression/ML-based fit applied to data
    (ex. MoCHI, Enrich2, etc.)"""
```

src/proteingym/base/assay.py
```diff
@@ -29,21 +29,129 @@
...
+
+
+class TargetPhenotype(StrEnum):
+    """TargetPhenotype
```
hredestig:
```suggestion
    """Target phenotype
```

src/proteingym/base/assay.py
```diff
@@ -29,21 +29,129 @@
...
+
+    Target phenotype the assay attempts to capture.
+    This language will probably be updated soon.
```
hredestig:
```suggestion
```

Everything is subject to change and suggest to avoid the 90s style 'website under construction' remarks (for those old enough to remember 😅)

src/proteingym/base/assay.py
```diff
@@ -253,6 +361,25 @@ class AssayManifestSection(_ManifestSection):
...
+
+    transformation: AssayTransformation | None = None
+    """The transformation applied to the raw assay data."""
+
+    target_phenotype: TargetPhenotype | None = None
+    """The property the assay attempts to capture."""
+
+    has_uncertainty: bool | None = None
+    """Authors include some form of assay uncertainty measure (they have replicate
+    experiments and/or correlation to low-throughput data)"""
```
hredestig:
```suggestion
    """Has uncertainty
    
    Authors include some form of assay uncertainty measure (they have replicate
    experiments and/or correlation to low-throughput data)"""
```

src/proteingym/base/assay.py
```diff
@@ -457,6 +584,25 @@ class Assay(AssayRaw):
...
```
hredestig:
```suggestion
    """Has uncertainty
    
    Authors include some form of assay uncertainty measure (they have replicate
    experiments and/or correlation to low-throughput data)"""
```

src/proteingym/base/assay.py
```diff
@@ -473,6 +619,11 @@ def target_feature_names(self) -> list[str]:
         non_target_names = [f.name for f in self.non_targets]
         return [name for name in all_names if name not in non_target_names]
 
+    @property
+    def number_of_records(self) -> int:
```
hredestig:
```suggestion
    def number_of_variants(self) -> int:
```
oops - i meant the original implementation was more like number of records, but your updated is  inline with number of unique variants..

hredestig (changes requested):
A couple of more changes in return - thanks for patience with the pedantisms. 
```
"""Title

Long description
"""
```
...but i'll look into adding linting support for this

src/proteingym/base/sequence.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

src/proteingym/base/assay.py
ethane4: see new commit

hredestig (approved):
Great, Thanks for the patience!

PR #467, opened 2026-05-15
replacing evedesign msa w/ Biotite msa

ethane4:
# Changes

This PR removes the evedesign MSA/Sequences dependency from pg-base and uses biotite.sequence.io.fasta.FastaFile in its place. The FastaFile object is pretty versatile and will allow people to do pretty much whatever they want with it downstream.  

Questions/points for discussion:
- I didn't add some global dataset validation function yet (ex. check MSA query matches dataset sequence, structure sequence, etc.), we may need to think about how to handle datasets that would--for whatever reason--have more than one sequence in the sequences portion of the manifest
- In evedesign, I will probably write the from_pgbase to work with a pgbase manifest object (to avoid loading in the msa/structures/etc.), which means the dataset object in this repo can do pretty much whatever we want it to wrt loading in/validating data etc.

# Checklist

- [ ] I broke the PR down so that it contains a reasonable amount of changes for an effective review
- [ ] I performed a self-review of my code. Amongst other things, I have commented my code in hard-to-understand areas.


src/proteingym/base/msa.py
```diff
@@ -266,7 +262,7 @@ def from_manifest_section(
             NotImplementedError if the file type is not supported.
         """
         name = section.name or section.path.stem
-        value = Sequences.from_file(section.path, format=section.format.value)
+        value = fasta.FastaFile.read(section.path)
```
karel-w: We should discuss what we actually want to output as the MSA. @hredestig @thomashopf @florisvdf 

Returning a fastafile gives us an unaligned dictionary, where each key is cryptic (as the keys in this case are taken from the a3m header line, which can be a lot of different things depending on the a3m generation software). So you'll use the .values() anyway, so why then not assign the .values() as the MSA?

I think it would make sense to do a list[Seqs], we could use the same Seq object as in our Sequences (Bio.Seq) but would add in another biopython dependency. If we want to keep with Biotite we could return a Alignment object (https://www.biotite-python.org/latest/apidoc/biotite.sequence.align.Alignment.html#) 

But i think it will be nicest to actually return a list[Seq(aligned_sequence)] or list[str(aligned_sequence] when calling msa.value

What is actually the input that people really use for their MSA based models? Just a list of aligned sequences?

I think we also need to provide an msa.file option such that any model that has its own msa read method we can pass the file too (e.g. as bytes).

karel-w (commented):
Let's provide some representation of the aligned sequences (set, list, w/e) but atleast provide back the aligned sequences so that its easy to use. 

Unfortunately Biotite does not have nice MSA support (why I opted for Biopython).
We could choose to return a list[Seq] objects, however this breaks for insertions. Lets discuss best possibility and actual current use in MSA based models.

src/proteingym/base/msa.py
ethane4: MSA inputs are pretty variable for each model (ex. the original PoET takes a giant string/concatenated list of sequences w/ start/stop characters, AF2 takes a list of strings w/ insertions stripped, etc.). About half of the MSA-based models to take the insertion information into account (ex. AF3 stores a mean insertion count per position, some models don't explicitly use an 'aligned' alignment - just a collection of seqs). 

My thought was to have the actual .value attribute in the pgbase dataset be as high-level as possible (so people can just do whatever they want with it downstream), but in principle it would be easy to create a list of aligned strings ex.

aln = fasta.get_a3m_alignments(fasta.FastaFile.read(section.path), seq_type=ProteinSequence)
list_of_strings_with_insertions_stripped = aln[0][aln[0].trace[:, 0] != -1].get_gapped_sequences()

Is there an advantage to storing the individual sequences as Bio.Seq objects instead of plain strings? Another thing to note is that converting a biotite FastaFile to an alignment object makes all of the sequence keys/headers inaccessible...

florisvdf: I think having access to the sequence headers/keys is such a fringe case that they probably don't need to be included in msa.value. Users can always resort to msa.file if they need them. 

Storing the sequences as biotite `Sequence`s has the added advantage of automatic validation for alphabet membership, but I'm not aware if any other advantages. @karel-w seems like a list of strings would be fine?

Also somewhat related, I noticed that biotite's `get_a3m_alignments` silently casts invalid residues denoted by O to K, and that it doesn't support certain non-canonical amino acids like selenocysteine, which is something to be aware of when using biotite and good to check for during validation.

ethane4: Ok seems like the best option is to just add the lines above and set value = list_of_strings_with_insertions_stripped?

I suppose the one thing this doesn't cover would be validating that the sequences only contain the 20 canonical amino acids--even if we wanted to support noncanonicals most pretrained models don't support them. Could always just add this check to MSA validation

hredestig: I see several points here coming up here

- BIopython or biotite? Let's decide on one of these. #464 
- What is the `.value` of `MSA`? 
  - Data structure is necessary, at least temporary, to enable validation of the manifest
    - Suggest to create a table of some example models in scope and detail what would result in the least amount of boilerplate for them to use pgdata. And write an ADR to document this. 
  - Raw file (content) also necessary to enable downstream consumers that are built to deal with files
- Non-canonical residues: we have `SequenceAlphabet` that could be used for validation.

karel-w:
> One point of discussion - for MSA writing, we assemble a biotite FastaFile object with arbitrary headers (ex. seq_1, seq_2, ...) and use the FastaFile.write() method to create the a3m. Since we are going the route of stripping gaps and changing the insertions to uppercase in msa.value, the alignment basically gets destroyed when you archive a dataset.

Why do we want to strip gaps and change insertions? I thought the goal of the move to A3M is to keep this information?

src/proteingym/base/msa.py
```diff
@@ -17,6 +18,60 @@
 )
 
 
+class PSequence(Sequence):
```
hredestig: Can we get a docstring here please like including explanation  why we aren't we just using https://www.biotite-python.org/latest/apidoc/biotite.sequence.ProteinSequence.html  (to support gap and  lowercase). And more descriptive name please, like `MsaProteinSequence`

Also, the long tuple could be written as 

```python
residues = "ACDEFGHIKLMNPQRSTWVY"
return LettterAlphabet(tuple(residues) + tuple(residues.lower() + "-")
```
for brevity

tests/test_dataset.py
```diff
@@ -203,7 +203,7 @@ def test_list_datasets_json_serialization() -> None:
         value=AtomArray(10),
     )
 
-    msa = MSA(name="test", value=Sequences([]))
```
hredestig, line 206: value indicated to be `list[PSequence]`?

tests/test_msa.py
```diff
@@ -111,41 +111,53 @@ def test_msa_manifest_section_serialize_path_as_posix_relative_to(
 def test_msa_minimal() -> None:
     """Only name and value are required for a minimal MSA."""
     try:
-        MSA(name="test", value=Sequences([]))
-    except ValidationError as e:
```
hredestig, line 115: why not `ValidationError`

hredestig (changes requested):
Just a few minor comments

src/proteingym/base/msa.py
ethane4: swapped + added X character to alphabet (adding X to the alphabet might be worth raising an issue over to discuss)

tests/test_dataset.py
ethane4: swapped

tests/test_msa.py
ethane4: swapped

hredestig: But I meant - why not keep it as it was? Are we getting a different exception now than `ValidationError` if so, let's catch that? Either way the tests are typically expected to raise `AssertionError`

hredestig (approved):
Great, Just a discretionary final comment here

tests/test_msa.py
ethane4: you are suggesting change it back to an AssertionError ?

src/proteingym/base/msa.py
```diff
@@ -266,8 +280,12 @@ def from_manifest_section(
             NotImplementedError if the file type is not supported.
         """
         name = section.name or section.path.stem
-        value = Sequences.from_file(section.path, format=section.format.value)
+        a3m = fasta.FastaFile.read(section.path)
+        seq_iter = iter(a3m.values())
```
florisvdf, line 284: Why do we need `iter` here? Isn't `a3m.values()` already an iterable?

tests/test_msa.py
```diff
@@ -111,41 +111,68 @@ def test_msa_manifest_section_serialize_path_as_posix_relative_to(
...
     """An MSA can be created from a manifest section with A3M file."""
     section = MSAManifestSection(path=a3m_file, format=MSAFormat.A3M)
 
     msa = MSA.from_manifest_section(section)
 
     assert msa.name == "structure"
-    assert isinstance(msa.value, Sequences)
+    assert isinstance(msa.value, list)
+    assert all(isinstance(s, MsaProteinSequence) for s in msa.value)
+    assert msa.value == msa_value
```
florisvdf, line 160: Not immediately clear why these should be the same. I see that both the msa_value and multiple_sequence_alignment fixture have the same sequences defined in them, so it works, but maybe better to infer both fixtures from a fixture of the sequences. Don't have a very strong opinion on this though, I'll leave this up to you if you want to make that change or not.

florisvdf (commented):
Nice, left two minor comments

src/proteingym/base/msa.py
ethane4: I never explicitly checked if it was an iterable, but I was just going directly off the biotite source code and for some reason theydid the same for the FastaFile obj when converting to sequences

tests/test_msa.py
ethane4: Probably fine to just leave as is?

florisvdf (approved):

PR #471, opened 2026-05-20
Adding the QuantileSplitter

florisvdf:
# Changes

Adds the `QuantileSplitter` and a new fixture for testing splits: `dataset_with_realistic_assays`.

The `QuantileSplitter` is intended to reserve a fraction of hit variants for the test set. Hit variants are defined as variants with target values that exceed the target values of all samples in the training set. The threshold for classifying variants as "hit variant" is determined by the `quantile` parameter.  

The `QuantileSplitter` divides a dataset into a lower, and an upper target quantile. The `fraction` parameter then determines what fraction of the lower quantile will make up the training data. The test data is then composed by uniformly sampling `1 - fraction` from the lower quantile and `1 - fraction` from the upper quantile. 

It's important to note that since the quantile split is dependent on the target values, the split is currently only defined for a single target. Additionally, this version currently only splits the data into a train and test set. There is no support yet for additional sets.

**Points of discussion**

1. I've noticed that all previously implemented splits share the same signature. Is that something that we want to enforce? If so, it will require a bit of refactoring of the `QuantileSplitter`. Perhaps then also warrants addition of an ABC `Splitter` class.
2. To properly test the `QuantileSplitter`, I've created a new fixture `dataset_with_realistic_assays`. The previous dataset fixtures with assay data have mostly constant target values. Is there a reason for this? If not, should we update them to have varied target distributions?

# Checklist

### DIFF
--- a/src/proteingym/base/splits.py
+++ b/src/proteingym/base/splits.py
@@ -104,20 +104,14 @@ def _subsample_mask(
     """
     Subsample True values in a boolean mask.
 
-    Parameters
-    ----------
-    mask : npt.NDArray
-        Boolean array to subsample
-    fraction : float
-        Fraction of True values to keep (0.0 to 1.0)
-        E.g., 0.8 means keep 80% and flip 20% to False
-    random_state: Seed or random state for
-        reproducibility. If None, the global numpy random state is used.
+    args
+        mask: Boolean array to subsample
+        fraction: Fraction of True values to keep (0.0 to 1.0)
+            E.g., 0.8 means keep 80% and flip 20% to False
+        random_state: reproducibility. If None, the global numpy random state is used.
 
-    Returns
-    -------
-    npt.NDArray
-        New boolean mask with subsampled True values
+    Returns:
+        npt.NDArray: New boolean mask with subsampled True values
     """
     if not 0 <= fraction <= 1:
         raise ValueError("Fraction must be between 0 and 1")
--- a/src/proteingym/base/splits.py
+++ b/src/proteingym/base/splits.py
@@ -234,9 +234,8 @@ class QuantileSplitter:
         for assay in dataset.assays:
             target_names_in_assay = [e.name for e in assay.fields]
             if target not in target_names_in_assay:
-                columns = []
-                train_assay_slices.append(AssaySlice(records=None, columns=columns))
-                test_assay_slices.append(AssaySlice(records=None, columns=columns))
+                train_assay_slices.append(AssaySlice(records=None, columns=[]))
+                test_assay_slices.append(AssaySlice(records=None, columns=[]))
             else:
                 columns = [assay.sequence_feature_name, target]
                 target_index = next(
--- a/src/proteingym/base/splits.py
+++ b/src/proteingym/base/splits.py
@@ -13,6 +13,7 @@ import logging
 import numbers
 
 import numpy as np
+import numpy.typing as npt
 from Bio.Seq import Seq
 
 from .assay import AssaySlice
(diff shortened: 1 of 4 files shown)


src/proteingym/base/splits.py
```diff
@@ -95,6 +96,35 @@ def _sequences_to_mask(selection: list[Seq], *, all_sequences: list[Seq]) -> lis
     return mask
 
 
+def _subsample_mask(
+    mask: npt.NDArray,
+    fraction: float,
+    random_state: int | np.random.RandomState | None = None,
+) -> npt.NDArray:
+    """
```
hredestig, line 104: Nitpick (we should have a linter for this) - docstrings are
```python
"""Title / Description"""
```
or if longer like in this case
```python
"""Title

Long description
"""
```

src/proteingym/base/splits.py
```diff
@@ -95,6 +96,35 @@ def _sequences_to_mask(selection: list[Seq], *, all_sequences: list[Seq]) -> lis
...
+    Subsample True values in a boolean mask.
+
+    args
```
hredestig: Args:

src/proteingym/base/splits.py
```diff
@@ -146,6 +176,83 @@ def _unique_sequences_for_targets(
     return list(sequences)
 
 
+class QuantileSplitter:
+    """Split a dataset a random training set and a test set with a sample of hit
```
hredestig: docstring format like above

src/proteingym/base/splits.py
```diff
@@ -146,6 +176,83 @@ def _unique_sequences_for_targets(
...
+    variants defined as values exceeding a quantile threshold. This split is only
+    defined for a single target in the dataset, because the quantile threshold can only
+    be defined for a single target.
+    Args:
```
hredestig, line 191: empty line

src/proteingym/base/splits.py
```diff
@@ -146,6 +176,83 @@ def _unique_sequences_for_targets(
...
+
+        Returns:
+            Subsets: The subsets containing the splits.
+        """
+        train_assay_slices = []
+        test_assay_slices = []
+        for assay in dataset.assays:
+            target_names_in_assay = [e.name for e in assay.fields]
+            if target not in target_names_in_assay:
+                columns = []
```
hredestig: pointless assignment, just 
```python
                train_assay_slices.append(AssaySlice(records=None, columns=[]))
```

src/proteingym/base/splits.py
```diff
@@ -146,6 +176,83 @@ def _unique_sequences_for_targets(
...
+                train_assay_slices.append(AssaySlice(records=None, columns=columns))
+                test_assay_slices.append(AssaySlice(records=None, columns=columns))
+            else:
+                columns = [assay.sequence_feature_name, target]
+                df = assay.to_df(target_names=[target])
```
hredestig: slightly more efficient would be  something like
```python
target_index = assay.fields.index(target)
np.array(r[target_index] for r in assay.records)
```
since you would avoid to create the dataframe which is then never used

tests/conftest.py
```diff
@@ -210,6 +210,74 @@ def dataset_with_assays() -> Dataset:
...
+        )
+        for i, s in enumerate(
+            ["AA", "CC", "DD", "EE", "FF", "GG", "HH", "II", "JJ", "KK"]
+        )
+    ]
+    assay1 = Assay(
+        name="assay2",
+        records=[
+            (sequences[0], 0.12, 1.83),
+            (sequences[0], 0.47, 1.21),  # duplicate sequence to handle that
```
hredestig, line 231: makes sense - although at some point we agreed that in `Assay` we don't have duplicate sequences. It should be a pre-processing responsibility to from `AssayRaw` with duplicates to `Assay` with summarized data.

tests/conftest.py
```diff
@@ -210,6 +210,74 @@ def dataset_with_assays() -> Dataset:
     return dataset
 
 
+@pytest.fixture
+def dataset_with_realistic_assays() -> Dataset:
```
hredestig: realistic is maybe a big word here :) `dataset_with_varying_targets`

hredestig (changes requested):
Looks good, just some minor nitpicks

tests/conftest.py
florisvdf: Ok. I have left this unchanged for now to mirror `dataset_with_assays`.

hredestig (approved):

PR #474, opened 2026-06-09
Adding the KFoldQuantileSplitter

florisvdf:
# Changes

Adding K-fold cross validation support for quantile splitting. 

## Newly introduced convention for Subsets structure

Since quantile splitting creates test folds that contain samples that never appear in the training data, the structure of the resulting `Subsets` object is slightly different from the `Subsets` object created by the `KFoldSplitter`. The former creates a dictionary with a key `train_folds` storing all the training sets for each fold, and a key `test_folds` storing all the test folds. 

Existing `proteingym-benchmark` models do not yet comply with this structure.

# Checklist


src/proteingym/base/splits.py
```diff
@@ -266,6 +311,126 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+                train_assay_slices[fold].append(
+                    AssaySlice(records=train_mask.tolist(), columns=columns)
+                )
+                test_assay_slices[fold].append(
+                    AssaySlice(records=test_mask.tolist(), columns=columns)
+                )
+
+        slices = defaultdict(list)
+        for fold in range(self.n_splits):
+            slices["train_folds"].append(DatasetSlice(assays=train_assay_slices[fold]))
```
hredestig: Naming the slices was meant for user to be able to store named slices like output of 'random'. This splitter deviates from that pattern by itself defining what the name should be.
```python
subsets = Subsets(dataset=dataset)

random_classic = RandomSplitter(fractions=[0.8, 0.1, 0.1])
random_kfold = RandomSplitter(fractions=[0.2, 0.2, 0.2, 0.2, 0.2])

subsets.update(classic=random_classic.split(dataset=dataset))
subsets.update(kfold=random_kfold.split(dataset=dataset))
```
now subsets have keys `classic` and `kfold`

but with this we can do the same because
```python
subsets = Subsets(dataset=dataset)

random = RandomSplitter(fractions=[0.8, 0.1, 0.1])
kquant = KFoldQuantileSplitter(n_splits=5)

subsets.update(classic=random.split(dataset=dataset))
subsets.update(kfold=quant.split(dataset=dataset)) # ??
```

I think either we rethink this logic - or more inline would be to have something like

```python
dataset
subsets = Subsets(dataset=dataset)

kquant_train = KFoldQuantileSplitter(n_splits=5, purpose="train")
kquant_test = KFoldQuantileSplitter(n_splits=5, purpose="test")

subsets.update(k_quantile_train=kquant_train.split(dataset=dataset))
subsets.update(k_quantile_test=kquant_test.split(dataset=dataset))
```

hredestig (commented):
Let's think of the api a bit - and also, can you please add some content to the splits notebook as well to demonstrate how kfold-quantile is used

src/proteingym/base/splits.py
florisvdf: The `split` method of the `KFoldQuantileSplitter` now returns a tuple of two `Subsets` objects, each containing a list of `DatasetSlice` objects. One can now update the `Subsets` object like so:

```
dataset
subsets = Subsets(dataset=dataset)

splitter = KFoldQuantileSplitter(n_splits=5)
train_subsets, test_subsets = splitter.split(dataset)

subsets.update(k_quantile_train=train_subsets)
subsets.update(k_quantile_test=test_subsets)
```

docs/notebooks/05_Creating_Dataset_Splits.ipynb
karel-w: Add quantilesplit

docs/notebooks/05_Creating_Dataset_Splits.ipynb
```diff
@@ -411,6 +408,58 @@
     "    print(f\"  {split_names[enum]}: {count} samples ({percentage:.1f}%)\")"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "### Quantile Splits\n",
+    "Quantile splits hold out a fraction of high performing variants for the test set. It works by creating an upper and lower quantile of samples based on their property variants, and then sampling a fraction from the lower quantile to test the training set, and sampling a small fraction from both the lower and the upper quantile to form the test set.\n",
```
karel-w: > and then sampling a fraction from the lower quantile to test the training set

assume you mean

> and then sampling a fraction from the lower quantile to **form** the training set

docs/notebooks/NEIME_2019.splits.pgdata
karel-w, line 1: Why do we need the .splits.pgdata file for the notebooks?

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
-                    upper_mask,
-                    fraction=1 - self.fraction,
-                    random_state=self.random_state,
-                ) | (~train_mask & lower_mask)
-                train_assay_slices.append(
-                    AssaySlice(records=train_mask, columns=columns)
-                )
-                test_assay_slices.append(AssaySlice(records=test_mask, columns=columns))
+        assays_contain_target = _assays_contain_target(dataset, target)
+        if not any(assays_contain_target):
```
karel-w, line 396: If the split is empty, should we raise an error instead of propagating an empty split?

src/proteingym/base/splits.py
```diff
@@ -178,6 +220,112 @@ def _unique_sequences_for_targets(
...
+    variable_names = [variable.name for variable in dataset.assay_variables]
+    combinations = {
+        tuple((name, assay.variables.get(name)) for name in variable_names)
+        for assay, contains_target in zip(
+            dataset.assays, assays_contain_target, strict=True
+        )
+        if contains_target
+    }
+    if len(combinations) > 1:
+        raise ValueError(
```
karel-w, line 258: Raise which assay variables are providing the error

src/proteingym/base/splits.py
```diff
@@ -178,6 +220,112 @@ def _unique_sequences_for_targets(
...
+    Returns:
+        int: The number of high-property variants in the test slice.
+    """
+    df = dataset[test_slice].to_df(target_names=[target])
+    if df.is_empty():
+        return 0
+    return int((df[target] > threshold).sum())
+
+
 class QuantileSplitter:
```
karel-w, line 330: Why do we only do two defined slices? Should atleast add a validation slice or allow users to set e.g. 2 or 3 slices with sizes.

src/proteingym/base/splits.py
```diff
@@ -178,6 +220,112 @@ def _unique_sequences_for_targets(
...
+        if contains_target
+    }
+    if len(combinations) > 1:
+        raise ValueError(
+            "Quantile splitting does not support combining assays with varying assay "
+            "variables."
+        )
+
+
+def _target_assay_slices(
```
karel-w, line 265: Can currently return empty assay slices if assays_contain_target is [false, false, ..], should we not always want to make sure we atleast return a slice with some information?

src/proteingym/base/splits.py
karel-w, line 339: add `fraction` arg to the docstring

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+            dataset, assays_contain_target=assays_contain_target
+        )
+        df = dataset.to_df(target_names=[target])
+        if not df[target].dtype.is_numeric():
+            raise ValueError(
+                f"KFoldQuantileSplitter requires a numeric target, but '{target}' is "
+                f"{df[target].dtype}."
+            )
+        target_values = df[target].to_numpy()
+        sequences = [Seq(s) for s in df[SEQUENCE].to_list()]
```
karel-w: Do we need .to_list()? df[SEQUENCE] should already be an iterable?

src/proteingym/base/splits.py
```diff
@@ -178,6 +220,112 @@ def _unique_sequences_for_targets(
...
+        dataset: The dataset that is sliced.
+        test_slice: The test slice to inspect.
+        target: Target field name on which the threshold is defined.
+        threshold: The quantile threshold separating low- and high-property variants.
+
+    Returns:
+        int: The number of high-property variants in the test slice.
+    """
+    df = dataset[test_slice].to_df(target_names=[target])
+    if df.is_empty():
```
karel-w, line 325: Should we work with empty test slices?

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+            selected_sequences=test_sequences,
+            assays_contain_target=assays_contain_target,
+        )
+
         train_dataset_slice = DatasetSlice(assays=train_assay_slices)
         test_dataset_slice = DatasetSlice(assays=test_assay_slices)
+        top_k = _count_high_property_variants(
+            dataset, test_dataset_slice, target=target, threshold=threshold
+        )
+        test_dataset_slice = DatasetSlice(
```
karel-w, line 447: Do we need to create a test_dataset_slice twice to obtain the count of high property variants?

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+
+    def __init__(
+        self,
+        quantile: float,
+        n_splits: int,
+        *,
+        shuffle: bool = False,
+        random_state: int | np.random.RandomState | None = None,
+    ) -> None:
+        if not 0 <= quantile <= 1:
```
karel-w: What happens if quantile is 0 or 1? We get empty slices?

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+            dataset: The dataset to split.
+            target: Target field name to include in the splits.
+
+        Returns:
+            Subsets: The subsets containing the splits.
+        """
+        assays_contain_target = _assays_contain_target(dataset, target)
+        train_slices = []
+        test_slices = []
+        if not any(assays_contain_target):
```
karel-w, line 523: Do we want to provide empty splits?

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+
+        The two resulting Subsets stores their slices as lists of folds with paired
+        ordering. The former contains all training folds, and the latter all tests
+        folds.
+
+        Args:
+            dataset: The dataset to split.
+            target: Target field name to include in the splits.
+
+        Returns:
```
karel-w, line 517: Returns a tuple of Subsets

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+        if n_splits < 2:
+            raise ValueError("Number of splits must be at least 2.")
+        if not shuffle and random_state is not None:
+            logger.warning("random_state is ignored when shuffle is False.")
+        self.quantile = quantile
+        self.n_splits = n_splits
+        self.shuffle = shuffle
+        self.random_state = _check_random_state(random_state)
+
+    def split(self, dataset: Dataset, *, target: str) -> tuple[Subsets, Subsets]:
```
karel-w, line 499: Both split methods share very similar logic:

- checks if assay_contains_target
- checks for empty targets
- check for single variable (assay conditions)
- check df and numeric type
- computes treshold.

Could be done in a helper method

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+        """Splits the dataset into a Subsets object storing k training and test sets.
+
+        For a single target, the quantile threshold is calculated based on
+        self.quantile. The threshold is used to divide the data into an upper and lower
+        interval. Both intervals are split into self.n_splits folds. For fold i, the
+        test set is composed of fold i of the upper interval and fold i of the lower
+        interval, while the training set is composed of every fold except fold i from
+        only the lower interval.
+
+        The two resulting Subsets stores their slices as lists of folds with paired
```
karel-w: very pedantic, non plural

> The two resulting Subsets store their slices...

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+            logger.warning("random_state is ignored when shuffle is False.")
+        self.quantile = quantile
+        self.n_splits = n_splits
+        self.shuffle = shuffle
+        self.random_state = _check_random_state(random_state)
+
+    def split(self, dataset: Dataset, *, target: str) -> tuple[Subsets, Subsets]:
+        """Splits the dataset into a Subsets object storing k training and test sets.
+
+        For a single target, the quantile threshold is calculated based on
```
karel-w, line 502: > For fold i, the test set is composed of fold i of the upper interval and fold i of the lower interval, while the training set is composed of every fold except fold i from only the lower interval.

This describes that the training set `i` is composed of every fold except `i_lower`, does this mean that `i_upper` is in the training set?

src/proteingym/base/splits.py
```diff
@@ -226,43 +387,216 @@ def split(self, dataset: Dataset, *, target: str) -> Subsets:
...
+
+        Returns:
+            Subsets: The subsets containing the splits.
+        """
+        assays_contain_target = _assays_contain_target(dataset, target)
+        train_slices = []
+        test_slices = []
+        if not any(assays_contain_target):
+            for _ in range(self.n_splits):
+                empty = [AssaySlice(records=None, columns=[]) for _ in dataset.assays]
```
karel-w: also we create the empty slice n times, could create once and assign within the loop

tests/test_splits.py
```diff
@@ -364,6 +365,186 @@ def test_quantile_splitter_splits_with_target_not_in_all_assays(
...
+
+def test_kfold_quantile_splitter_splits_are_disjoint(
+    dataset_with_assays: Dataset,
+) -> None:
+    """Test that KFoldQuantileSplitter splits are disjoint."""
+    splitter = KFoldQuantileSplitter(quantile=0.75, n_splits=2)
+    train_subsets, test_subsets = splitter.split(
+        dataset_with_assays, target="DMS Score"
+    )
+    for train, test in zip(train_subsets, test_subsets, strict=True):
```
karel-w, line 432: Should we test the records or sequences instead of the complete subset? e.g.

```
train_sequences = {
    str(record[0].value)
    for assay in train.assays
    for record in assay.records
}
```

Then you can compare two dictionaries with train_sequences.isdisjoint(test_sequences)

tests/test_splits.py
```diff
@@ -364,6 +365,186 @@ def test_quantile_splitter_splits_with_target_not_in_all_assays(
...
+    original = set()
+    for assay in dataset_with_assays.assays:
+        target_idx = [f.name for f in assay.fields].index(target)
+        for record in assay.records:
+            original.add((str(record[0].value), record[target_idx]))
+
+    assert covered == original
+
+
+def test_kfold_quantile_splitter_test_folds_contain_hit_variants(
```
karel-w, line 468: pedantic, but this test tests if we contain hit variants and all hit variants are higher than train variants, so actually tests more than just if there are hit variants.

tests/test_splits.py
```diff
@@ -364,6 +365,186 @@ def test_quantile_splitter_splits_with_target_not_in_all_assays(
...
+            dataset_with_varying_targets.assays,
+            train_slice.assays,
+            test_slice.assays,
+            strict=True,
+        ):
+            if not test_assay_slice.columns:
+                continue  # Assay without the target is empty.
+            target_idx = [f.name for f in assay.fields].index(target)
+            all_values = np.array([r[target_idx] for r in assay.records], dtype=float)
+            threshold = float(np.quantile(all_values, quantile))
```
karel-w: Should this be done on the whole dataset (df) level instead of the assay? This is what you do in [splits:543](https://github.com/ProteinGym/proteingym-base/blob/53d7edb0dbfe00889ea5c1b59b4a6726af482488/src/proteingym/base/splits.py#L543)

tests/test_splits.py
```diff
@@ -364,6 +365,186 @@ def test_quantile_splitter_splits_with_target_not_in_all_assays(
...
+            train_mask = np.asarray(train_assay_slice.records, dtype=bool)
+            test_mask = np.asarray(test_assay_slice.records, dtype=bool)
+            train_values = all_values[train_mask]
+            hits = all_values[test_mask][all_values[test_mask] > threshold]
+            if train_values.size and hits.size:
```
karel-w: edge case, if we have no train_values and no hits (e.g. the case of empty slices), you will not reach the assert.

tests/test_splits.py
```diff
@@ -561,3 +742,261 @@ def test_predefined_splitter_assay_missing_split_column_yields_empty_view(
...
+    target = "DMS Score"
+    splitter = QuantileSplitter(quantile=0.75, fraction=0.5, random_state=0)
+    subsets = splitter.split(dataset_shared_sequence_across_assays, target=target)
+    _, test_slice = subsets.slices
+
+    test_df = dataset_shared_sequence_across_assays[test_slice].to_df(
+        target_names=[target]
+    )
+    aa_rows = test_df.filter(pl.col("sequence") == "AA")
+    assert all(value == 0.0 for value in aa_rows[target])
```
karel-w: You only test for value, not if its a hit or not. Might even be a hit variant since you extract it from the test slice?

tests/test_splits.py
```diff
@@ -561,3 +742,261 @@ def test_predefined_splitter_assay_missing_split_column_yields_empty_view(
...
+            dataset_with_varying_targets.to_df(target_names=[target])[
+                target
+            ].to_numpy(),
+            quantile,
+        )
+    )
+    expected = _aggregated_hits(
+        dataset_with_varying_targets, test_slice, target, threshold
+    )
+    assert test_slice.metadata["top_k"] == expected
```
karel-w, line 867:
```
assert expected > 0, "tests should have some hits"
assert test_slice.metadata["top_k"] == expected
```
Should cover the edge case of no hits.

tests/test_splits.py
```diff
@@ -561,3 +742,261 @@ def test_predefined_splitter_assay_missing_split_column_yields_empty_view(
...
+        )
+    )
+    for test_slice in test_subsets.slices:
+        expected = _aggregated_hits(
+            dataset_with_varying_targets, test_slice, target, threshold
+        )
+        assert test_slice.metadata["top_k"] == expected
+
+
+def test_kfold_quantile_splitter_top_k_total_equals_all_hits(
```
karel-w, line 1017: Should there be a test that each sequence is also appearing in exactly one fold?

karel-w (changes requested):
We now allow for empty splits, which will fail downstream when we cannot extract data from the split. I would suggest to fail early and raise errors when we obtain an empty split instead.

This also removes some of the edge cases in testing.

src/proteingym/base/splits.py
florisvdf: The `_count_high_property_variants` performs aggregation to determine the number of high property variants. For this we need a dataset slice. It doesn't look pretty this way but it is explicit and robust.

src/proteingym/base/splits.py
florisvdf: Yep you just get an empty training slice. True that it makes no sense to do that, but does it make more sense to have a quantile of 0.0001? What about 0.01, or 0.1. Don't know where to draw the line.

src/proteingym/base/splits.py
florisvdf: Yes but not identical. `KFoldQuantileSplitter` does fold iteration in this section. Not worth the added complexity having a helper function that accounts for that imo.

tests/test_splits.py
florisvdf: It's implemented this way to be consistent with other testing logic, e.g., `test_random_splitter_splits_are_disjoint`.

tests/test_splits.py
florisvdf: Semantics :p. I define a hit variant as a variant with a property value higher than what is in the training set. Thus checking for presence of hit variants in the test sit implies checking that those variants have property values exceeding those of the training variants.

tests/test_splits.py
florisvdf: Well spotted, thanks!

tests/test_splits.py
florisvdf: Thanks! Good point, fixed.

tests/test_splits.py
florisvdf: 👍

tests/test_splits.py
florisvdf: Yes, added.

src/proteingym/base/splits.py
karel-w: `if not 0 < quantile < 1`

florisvdf: But why is 0.000000001 a more valid parameter than 0?

src/proteingym/base/splits.py
florisvdf: Fair point. Now raises ValueError

src/proteingym/base/splits.py
florisvdf: Mostly because I wanted to keep things simple. I guess you could provide a list of quantiles to divide the data into multiple intervals, one for train, val, test (and optionally more). Definitely possible, but maybe not per se necessary for now? Can create an issue if this is needed in the future.

src/proteingym/base/splits.py
karel-w: 0 and 1 creates empty slices and should fail early, e.g. when selecting 0 or 1, we raise the ValueError of the Quantile not being between 0 and 1.

0.000001 would most likely be a misuse of the quantile selection, but e.g. in a 1.000.000 sized dataset 0.00001 is theoretically valid as you get atleast one sample in your quantile.

src/proteingym/base/splits.py
karel-w: Optionally more I guess you get KfoldQuantiles.

But I would strongly suggest to add the possibility for creating a validation dataset for the benchmarking. 

If we hypertune a model in the benchmarking suite, the validity of the performance and benchmarking on the test set is in question, as you are overfitting to the test.

Unless we want to say we only accept hypertuned models?

karel-w: And I would imagine the validation could be e.g. 80/20 split across the quantiles, then divide the 20% over both a test and validation set. 

E.g. doing a train/validation/test at 80/10/10 would make different quantiles which would be a slightly different method.

src/proteingym/base/splits.py
karel-w: We wouldn't expect to get an empty slice here, if the test slice is empty we should have failed earlier.

karel-w (approved):
Lets raise an issue for the validation of QuantileFolds to discuss this point, but in the meanwhile we can move on. 

Line 325 is now redundant since we would fail earlier when the slice is empty? Would maybe be a nice test to make sure _count_high_property_variants return correct counts.

src/proteingym/base/splits.py
florisvdf: Fair, updated to open interval

karel-w (approved):

PR #475, opened 2026-06-09
Adding metadata attributes to DatasetSlice and AssaySlice classes

florisvdf:
# Changes

Allows users to tag their slices with metadata. This is useful when users want to add:

- Splitting parameters
    - `fraction` for random splitting
    - `quantile` for quantile splitting
- Properties of the resulting slice
    - `top_k` for QuantileSplitting
- Annotation/description of slices created in usage other than splitting.

# Checklist

### DIFF
--- a/src/proteingym/base/assay.py
+++ b/src/proteingym/base/assay.py
@@ -453,7 +453,10 @@ class AssaySlice:
     """The boolean mask for the records. If None, all records are included."""
 
     metadata: dict[str, Union[str, float]] | None = None
-    """Metadata associated with the AssaySlice."""
+    """Metadata associated with the AssaySlice. The metadata can be used to describe
+    details about the slice, for instance properties of the slice, like the number of 
+    top performing variants contained in the assay slice, or any other information 
+    relevant to the user."""
 
     @classmethod
     def from_json(cls, contents: str) -> "AssaySlice":
--- a/src/proteingym/base/assay.py
+++ b/src/proteingym/base/assay.py
@@ -5,7 +5,7 @@ import json
 from collections.abc import Collection
 from enum import StrEnum
 from pathlib import Path
-from typing import Annotated
+from typing import Annotated, Union
 
 import annotated_types
 import polars as pl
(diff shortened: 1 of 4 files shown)


src/proteingym/base/assay.py
```diff
@@ -452,6 +452,9 @@ class AssaySlice:
     records: list[bool] | None = None
     """The boolean mask for the records. If None, all records are included."""
 
+    metadata: dict[str, Union[str, float]] | None = None
```
hredestig: Better written as `dict[str, str | float]`

src/proteingym/base/dataset.py
```diff
@@ -65,6 +65,9 @@ class DatasetSlice:
     assays: list[AssaySlice | list[bool | str]] | None = None
     """The list of assay slices. If None, all assays are included."""
 
+    metadata: dict[str, Union[str, float]] | None = None
```
hredestig: `str | float`

src/proteingym/base/assay.py
```diff
@@ -5,7 +5,7 @@
 from collections.abc import Collection
 from enum import StrEnum
 from pathlib import Path
-from typing import Annotated
+from typing import Annotated, Union
```
hredestig: python < 3.9 ..

src/proteingym/base/dataset.py
```diff
@@ -65,6 +65,9 @@ class DatasetSlice:
...
+    """Metadata associated with the DatasetSlice."""
```
hredestig: Elaborate on use-case for this?

src/proteingym/base/assay.py
```diff
@@ -452,6 +452,9 @@ class AssaySlice:
     records: list[bool] | None = None
     """The boolean mask for the records. If None, all records are included."""
 
+    metadata: dict[str, Union[str, float]] | None = None
+    """Metadata associated with the AssaySlice."""
```
hredestig: Elaborate please?

hredestig (commented):
I don't quite understand the use-case here - could we get some more explanation for how this is intended to be used? Is it convenience when storing for 'local / personal'  use or also for stored splits meant for public use? If the latter I wonder if we wouldn't benefit from using a structured class such that we clarified what e.g. 'k' means?

hredestig (approved):
Approved (but please let's merge #476 first 👮‍♂️😅)

PR #476, opened 2026-06-12
chore: add docstring linting

hredestig:
# Add linting to our docstrings

Well formatted docstrings make for tidiness and help ensuring webdocs get rendered correctly. And noone like to nitpick in PRs..

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,19 @@ filterwarnings = [
 line-length = 88
 
 [tool.ruff.lint]
-ignore = ["N805"]
+# Ignore rules that require docstrings to be present
+# Only check format/style of existing docstrings
+ignore = [
+    "N805",
+    "D100",  # Missing docstring in public module
+    "D101",  # Missing docstring in public class
+    "D102",  # Missing docstring in public method
+    "D103",  # Missing docstring in public function
+    "D104",  # Missing docstring in public package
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -70,7 +82,15 @@ select = [
     # "C", # flake8-comprehensions
     "B", # flake8-bugbear
     "A", # flake8-builtins
+    "D", # pydocstyle
 ]
 
+[tool.ruff.lint.pydocstyle]
+convention = "google"
```
florisvdf, line 89: Do we want to document motivation for using the google convention rather than something like numpy (which is apparently common in bioinformatics)?

florisvdf (approved):
This is new territory for me, but it looks nice. One question that came to mind is if we want to enforce some kind of docstring coverage with e.g. [pydocstring-coverage](https://pypi.org/project/pydocstring-coverage/#:~:text=Python%20docstring%20coverage&text=If%20the%20health%20of%20your,modules%20don't%20have%20docstrings).

pyproject.toml
florisvdf: Biotite uses this for instance. Though I gotta say that I think the google convention looks much more clean.

hredestig: I don't think there is any relevant motivation to give other than personal preference. Either is fine but I like google more.

PR #484, opened 2026-06-23
minor msa metadata change

ethane4:
# Changes

Resolves lack of Neff/L storage method

I think it would be good to have a Neff per L metadata attribute for MSAs in manifest. I also think it would be good to remove the num_significant attribute, considering it won't ever really be used + it is not something that we know the answer to at the time of manifest construction.

# Checklist


src/proteingym/base/assay.py
```diff
@@ -419,6 +419,11 @@ class AssayManifestSection(_ManifestSection):
     Authors include some form of assay uncertainty measure (they have replicate
     experiments and/or correlation to low-throughput data)"""
 
+    number_of_variants: int | None = None
```
hredestig, line 422: why do you need this in addition to the computed `Assay.number_of_variants` ?

src/proteingym/base/msa.py
```diff
@@ -94,8 +94,8 @@ class MSAMetadataManifestSection(BaseModel):
     )
     """Configuration for the Pydantic model."""
 
-    num_significant: int | None = None
-    """Number of evolutionary couplings that are considered significant."""
+    Neff_per_L: float | None = None
```
hredestig: style comment - attributes are lower-case so `neff_per_l`. I'm surprised ruff doesn't flag this 🤔

Also, is this not computable from the weights?

src/proteingym/base/assay.py
ethane4: I think it would be nice to have it displayed in the manifest itself as well

ethane4: (for fast data aggregation across manifests etc)

src/proteingym/base/msa.py
ethane4: swapped

ethane4: We could compute from weights but for very large MSAs I think this would make dataset generation take forever. Maybe not an issue for dealing with one dataset at a time

src/proteingym/base/assay.py
hredestig: Alright, that makes, sense - but then I guess we should add the computation of this (using `Assay.number_of_variants`) [here](https://github.com/ProteinGym/proteingym-base/blob/main/src/proteingym/base/dataset.py#L582) to make it populate it self upon save? And add info that this is computed if left empty. I think that would facilitate management and increase accuracy of the manifests.

src/proteingym/base/msa.py
hredestig: Same as the number of variants, we can
- add the attribute as you propose
- have it computed on dumping zip if still None at that time

That way, just leaving it missing will ensure you have a correct value and no need to compute first and then add the value manually when creating manifests. And, additional bonus is that the implementation for computing this would be in the repo for scrutiny

src/proteingym/base/assay.py
ethane4: Added some code to automatically compute number of variants

src/proteingym/base/msa.py
ethane4: Neff/L is quite an expensive calculation and I think the code would add a decent amount of overhead/extra dependencies (potentially)?

I can investigate potentially adding some method for this if we are really interested in a native Neff/L calculation fxn -- would also need to be tied to the theta value if provided

src/proteingym/base/assay.py
hredestig: I'm confused, you want it at the manifest level but the new addition is on assay?

I was thinking that in the `_create_manifest` method you would compute the number of variants  if still missing to ensure it is available when you have saved the manifest (if not already defined)?

ethane4: I think it is better to not trust the user and just force computation any time the assaymanifestsection is validated? _create_manifest will still call to the assay section validation fxns, so I think that the attribute would be recomputed anyways, I think?

hredestig: I agree - better to recompute always if computation is fast enough. Not sure how slow the neff_per_l can be but if significant, then only computing if missing would be a middle-ground

src/proteingym/base/msa.py
hredestig: I think adding the function to compute neff_per_l would be great - what's the definition, not just column sum of weights?

Can also do that in a different MR if you prefer

hredestig (approved):

src/proteingym/base/msa.py
ethane4: Created #489

src/proteingym/base/assay.py
ethane4: Will get around to this eventually, would probably add a numba dependency to pg-base if we wanted it to be fast #489

PR #487, opened 2026-06-24
Adding a predictions_delta method to the Dataset class for representing model predictions

florisvdf:
# Changes

Addresses https://github.com/ProteinGym/proteingym-benchmark/issues/173

Creates a minimal version of the original pg dataset for representing model predictions obtained by calling `dataset.predictions_delta(df, target)`, where `target` is the assay target that is predicted and `df` is a dataframe with columns `sequence` and the value of `target` containing model predictions. The method iterates over all assays in the dataset and for assays that share sequences with df AND have a target that matches the target argument, an updated assay is created with identical sequence records and predictions records storing predictions for the corresponding sequences, and null elsewhere.

The resulting `delta` pgdataset drops:
```
description
reference_sequence_name
assays_raw
sequences
structures
msas
msa_weights
publication
```

keeps:
```
assay_variables
```

and changes:

`name` to `"{name}_predictions"`
`assays` to `new_assays`
`assay_targets` to `next(t for t in self.assay_targets if t.name == target)` (Stripping all irrelevant targets)

# Checklist

### DIFF
--- a/src/proteingym/base/dataset.py
+++ b/src/proteingym/base/dataset.py
@@ -944,17 +944,6 @@ class Dataset(BaseModel):
         # Get the target field
         target_field = next(t for t in self.assay_targets if t.name == target)
 
-        # Sanitize target_field: empty string descriptions cause manifest validation
-        # errors, so convert them to None
-        if target_field.description == "":
-            target_field = dataclasses.replace(target_field, description=None)
-
-        # Sanitize assay_variables: same issue with empty descriptions
-        sanitized_variables = [
-            dataclasses.replace(v, description=None) if v.description == "" else v
-            for v in self.assay_variables
-        ]
-
         # Build prediction lookup from df (sequence string -> predicted value)
         # Use a dict for O(1) lookup per record
         predictions = {
@@ -1002,8 +991,7 @@ class Dataset(BaseModel):
                 predicted_value = predictions.get(seq_str)
                 new_records.append((sequence_obj, predicted_value))
 
-            # Build new fields: sequence + sanitized target field
-            # Use the sanitized target_field to ensure description consistency
+            # Build new fields: sequence + target field
             new_fields = [sequence_field, target_field]
 
             new_assays.append(
@@ -1021,7 +1009,7 @@ class Dataset(BaseModel):
                 "name": f"{self.name}_predictions",
                 "description": None,
                 "reference_sequence_name": None,
-                "assay_variables": sanitized_variables,
+                "assay_variables": self.assay_variables,
                 "assay_targets": [target_field],
                 "assays": new_assays,
                 "assays_raw": [],
@@ -1037,9 +1025,9 @@ class Dataset(BaseModel):
 def dummy_dataset() -> Dataset:
     """Create a trivial dataset for illustrative purposes."""
     fields = [
-        Field(name=SEQUENCE, description=""),
-        Field(name="numerical", description=""),
-        Field(name="categorical", description=""),
+        Field(name=SEQUENCE, description="Protein sequence field."),
+        Field(name="numerical", description="Numerical assay variable field."),
+        Field(name="categorical", description="Categorical assay variable field."),
     ]
     sequence1 = Sequence(
         name="seq1",
@@ -1067,10 +1055,10 @@ def dummy_dataset() -> Dataset:
     dataset = Dataset(
         name="dataset_with_single_assay",
         description="A dataset containing a single assay.",
-        assay_variables=[Field(name="var1", description="")],
+        assay_variables=[Field(name="var1", description="Assay variable.")],
         assay_targets=[
-            Field(name="numerical", description=""),
-            Field(name="categorical", description=""),
+            Field(name="numerical", description="Numerical assay measurement."),
+            Field(name="categorical", description="Numerical assay measurement."),
         ],
         assays=[assay],
         sequences=[],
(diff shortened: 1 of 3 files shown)


src/proteingym/base/dataset.py
```diff
@@ -867,13 +867,167 @@ def to_df(
...
+
+        IMPORTANT: Because this method matches records to predictions only using the
+        sequence string, this means that sequences measured under varying assay
+        variables will be assigned the same value. This mirrors the behavior from the
+        `to_df` method.
+
+        Args:
+            df: A DataFrame containing predictions. Must have:
+                - A ``sequence`` column with protein sequence strings
+                - A column named ``target`` with the predicted values
```
ethane4: This docstring might be inaccurate? (column named target)

florisvdf: Thanks! Fixed.

src/proteingym/base/dataset.py
```diff
@@ -867,13 +867,167 @@ def to_df(
...
+        target_field = next(t for t in self.assay_targets if t.name == target)
+
+        # Build prediction lookup from df (sequence string -> predicted value)
+        # Use a dict for O(1) lookup per record
+        predictions = {
+            row[0]: row[1] for row in df.select([SEQUENCE, target]).iter_rows()
+        }
+
+        # Warn if many predictions don't match any assay sequences
+        all_sequences = set()
```
hredestig, line 954: I wonder if not ValueError on the first missing sequence as default but with option to override that is better. I mean, either I expect to match (and if not, due to some mistake, i now get no indication of this. Or, I know they don't necessarily match, perhaps intentionally only 9%, then I get a spurious warning)

src/proteingym/base/dataset.py
```diff
@@ -867,13 +867,167 @@ def to_df(
...
+        """
+        # Validate target
+        valid_target_names = {t.name for t in self.assay_targets}
+        if target not in valid_target_names:
+            raise ValueError(
+                f"Target '{target}' is not a valid assay target. "
+                f"Valid targets: {', '.join(sorted(valid_target_names))}"
+            )
+
+        # Validate DataFrame columns
```
hredestig: style nitpick  but these comments are (imo) not helpful - claude loves to add them everywhere but the code itself is as easy to read as the comment so it doesn't add any value. I think comments are great when we do something a bit unusual, surprising, but e.g. to state that dict is O(1) not that newsworthy :D

src/proteingym/base/dataset.py
```diff
@@ -867,13 +867,167 @@ def to_df(
...
+                UserWarning,
+                stacklevel=2,
+            )
+
+        # Create new assays with predictions
+        new_assays = []
+        sequence_field = Field(name=SEQUENCE)
+
+        for assay in self.assays:
+            if target not in assay.target_feature_names:
```
hredestig, line 969: what's the use-case of these null records?

tests/test_dataset.py
```diff
@@ -126,6 +127,11 @@ def dataset_file(tmp_path: Path) -> Path:
     return dataset_path
 
 
+@pytest.fixture
```
hredestig: if function scoped, the fixture is just a rename of the dummy_dataset function

tests/test_dataset.py
```diff
@@ -263,3 +269,154 @@ def test_reference_sequence_not_present_errors() -> None:
...
+def test_predictions_delta_basic(dummy_example_dataset) -> None:
+    """Test basic predictions_delta functionality."""
+    # Create predictions DataFrame
+    predictions_df = pl.DataFrame(
+        {"sequence": ["ACDEFG", "GFEDCA"], "numerical": [1.2, 2.3]}
+    )
+
+    delta = dummy_example_dataset.predictions_delta(predictions_df, target="numerical")
+
+    # Check structure preservation
```
hredestig: bit of nitpick but in general it's better to avoid multiple asserts in tests. Here, you could craft the expected Dataset and then rely on the `__eq__` implementation to have a single assert. Makes test easier to read, and also easier to refactor

tests/test_dataset.py
florisvdf: This better?

`dummy_dataset = pytest.fixture(dummy_dataset)`

Or did you have something else in mind?

src/proteingym/base/dataset.py
florisvdf: This used to warn for extra predictions, i.e., predictions for sequences that can't be matched to records in the original dataset with the same target. Now throws a ValueError by default, which can be suppressed by the `allow_extra_predictions` flag.

src/proteingym/base/dataset.py
florisvdf: Removed the majority of these comments. I only left a few that I think improve legibility.

src/proteingym/base/dataset.py
florisvdf: Good point, I guess preserve the structure of the original assays, but if we're replacing both the field and records with nulls, may as well remove it entirely.

tests/test_dataset.py
florisvdf: Updated

src/proteingym/base/dataset.py
```diff
@@ -867,13 +867,161 @@ def to_df(
...
+
+        predictions = {
+            row[0]: row[1] for row in df.select([SEQUENCE, target]).iter_rows()
+        }
+
+        all_sequences = set()
+        for assay in self.assays:
+            all_sequences.update(str(record[0].value) for record in assay.records)
+        unused = set(predictions.keys()) - all_sequences
+        if unused:
```
hredestig: save an indent: `if unused and not allow_extra_predictions`

hredestig (approved):
Nice!

src/proteingym/base/dataset.py
florisvdf: Thanks, done

hredestig (approved):

PR #488, opened 2026-06-30
Feat/pyrefly (part I)

hredestig:
# Add pyrefly checking and fix some errors

This will take some time so I add type check scope a few modules at a time. First out are

```
project-includes = [
    # type checking is a work in progress
    "src/proteingym/base/__about__.py",
    "src/proteingym/base/__init__.py",
    "src/proteingym/base/__main__.py",
    "src/proteingym/base/assay.py",
    "src/proteingym/base/data_generators.py",
]
```

# Checklist

### DIFF
--- a/src/proteingym/base/assay.py
+++ b/src/proteingym/base/assay.py
@@ -25,7 +25,8 @@ from pydantic import (
 from .sequence import Sequence, SequenceAlphabet, SequenceType
 
 SEQUENCE = "sequence"
-RECORDS: TypeAlias = list[tuple[Sequence | str | int | float | bool | str | None, ...]]
+RECORD: TypeAlias = tuple[Sequence | str | int | float | bool | str | None, ...]
+RECORDS: TypeAlias = list[RECORD]
 
 
 class AssayFormat(StrEnum):
@@ -827,7 +828,7 @@ class Assay(AssayRaw):
         if n_recs == 0:
             lines.append("\t\t<no records>")
         for i, record in enumerate(self.records[:n_recs]):
-            seq = cast(Sequence, record[0])
+            seq = get_sequence(record)
             targets = record[1:]
             seq_str = str(seq.value)
             if len(seq_str) > 30:
@@ -902,7 +903,7 @@ class Assay(AssayRaw):
         if self.is_empty():
             sequence_alphabet = SequenceAlphabet.UNDEFINED
         else:
-            sequence_alphabet = cast(Sequence, self.records[0][0]).alphabet
+            sequence_alphabet = get_sequence(self.records[0]).alphabet
         return AssayManifestSection(
             name=self.name,
             description=self.description,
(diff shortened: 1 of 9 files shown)


src/proteingym/base/assay.py
```diff
@@ -273,7 +274,7 @@ def to_dict(self):
 
     # noinspection PyTypeChecker
     @functools.cached_property
-    def polars_type(self) -> pl.DataType:
+    def polars_type(self) -> type[pl.DataType]:
```
karel-w, line 277: Is pl.DataType not a type?

src/proteingym/base/assay.py
```diff
@@ -310,7 +311,7 @@ class _ManifestSection(BaseModel):
     description: str | None = None
     """A brief description"""
 
-    @field_validator("path", mode="before", check_fields=True)
+    @field_validator("path", mode="before", check_fields=True)  # noqa
```
karel-w, line 314: Why noqa on the field validators?

src/proteingym/base/assay.py
```diff
@@ -825,7 +841,8 @@ def __repr__(self) -> str:
         return "\n".join(lines)
 
     @classmethod
-    def from_manifest_section(cls, section: AssayManifestSection) -> "Assay":
+    # pyrefly: ignore [bad-override]
```
karel-w, line 844: Don't fully understand why these two cases are bad-override.

src/proteingym/base/assay.py
```diff
@@ -908,12 +926,12 @@ def as_manifest_section(self, *, path: Path) -> AssayManifestSection:
         )
 
     def to_df(
-        self, *, target_names: Collection[str] | str | None = None
+        self, *, field_names: Collection[str] | str | None = None
```
karel-w, line 929: Not sure I like the change from `target_names` -> `field_names`. 

`Dataset.to_df(target=activity)` is more explicit than `Dataset.to_df(field=activity)`

In our manifests we also call them assay.targets.

karel-w (approved):
LGTM, had a few questions for my own understanding.

src/proteingym/base/assay.py
hredestig: annotation `pl.DataType` implies you return an instance but the intention is to return the type so this is just a typo.

src/proteingym/base/assay.py
hredestig: because it is expected that classmethod is the outermost decorator. Except it mustn't be here.

src/proteingym/base/assay.py
hredestig: parent class takes `AssayRawManifestSection` and returns `AssayRaw`. Subclasses should respect that contract and take / return same type objects. Or you violate the important Liskov Substitution principle. Here though, I think that violation is preferrable to the alternative, which would be to have non-inherited `from_assay_raw_manifest_section` / `from_assay_manifest_section` -- so whatever with Liskov and his principle (here)

src/proteingym/base/assay.py
hredestig: Yes - but again, then the `to_df` is a bad override since you suddenly change the parameter names in a subclass. And, a target is a field so why do we need to change the name. In `Assay` we don't call them targets but just `fields`. And you still do `Dataset.to_df(target_names=...)` (unchanged) it's in `Assay.to_df` we have the change.

PR #490, opened 2026-07-01
Feat/pyrefly (part II)

hredestig:
# Fix remaining typing errors

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dependencies = [
     "scikit-learn>=1.5.0,<2.0.0",
     "pandas>=2.3.3",
     "requests>=2.32.3",
+    "pandas-stubs~=3.0.0",
 ]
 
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "scikit-learn>=1.5.0,<2.0.0",
     "pandas>=2.3.3",
     "requests>=2.32.3",
+    "pandas-stubs~=3.0.0",
```
karel-w, line 24: What is this for?

karel-w (dismissed):
LGTM, although I don't understand why we use pandas-stubs?

pyproject.toml
hredestig: pyrefly wants it to check types apparently

karel-w (approved):

PR #491, opened 2026-07-03
Add metrics module for benchmark evaluation

karel-w:
Migrate metric calculation, plugin-style discovery, and scoring helpers from proteingym-benchmark into base so the logic is testable and shared, keeping DVC pipeline glue in the benchmark repo. Exposes evaluate, calculate_selected_metrics, and calculate_metrics_by_mode as public API.

Resolves 204 in Benchmark

# Checklist

### DIFF
--- a/src/proteingym/base/metrics.py
+++ b/src/proteingym/base/metrics.py
@@ -10,26 +10,8 @@ datasets or specific cross-validation folds. All metrics use a plugin-style
 architecture: any function with the ``metric_`` prefix is automatically discovered
 and made available for calculation.
 
-Adding Custom Metrics:
-    To add a new metric, define a function following this pattern::
-
-        def metric_<name>(
-            ground_truth: Subsets | Dataset,
-            predicted: Dataset,
-            target: str,
-            split: str | None = None,
-            fold: int | None = None,
-        ) -> float:
-            scoring_df = prepare_and_validate_scoring_df(
-                ground_truth, predicted, target, split, fold
-            )
-            gt_values = scoring_df[target].to_numpy()
-            pred_values = scoring_df[f"{target}_pred"].to_numpy()
-            return custom_calculation(gt_values, pred_values)
-
-    The metric function is automatically discovered and made available for
-    calculation. The function name after ``metric_`` becomes the metric name used
-    in the ``selected_metrics`` argument.
+See the [Metrics documentation](../metrics.md) for a guide on how to design and
+add new metrics.
 """
 
 import inspect
--- a/src/proteingym/base/metrics.py
+++ b/src/proteingym/base/metrics.py
@@ -1,0 +1,643 @@
+"""Metric calculation for ProteinGym benchmark evaluation.
+
+This module provides a flexible, extensible framework for calculating performance
+metrics for machine learning models on protein datasets. It uses dynamic metric
+discovery to automatically detect and execute custom metric functions.
+
+Metrics operate on :class:`~proteingym.base.dataset.Dataset` or
+:class:`~proteingym.base.dataset.Subsets` objects, enabling evaluation on complete
+datasets or specific cross-validation folds. All metrics use a plugin-style
+architecture: any function with the ``metric_`` prefix is automatically discovered
+and made available for calculation.
+
+Adding Custom Metrics:
+    To add a new metric, define a function following this pattern::
+
+        def metric_<name>(
+            ground_truth: Subsets | Dataset,
+            predicted: Dataset,
+            target: str,
+            split: str | None = None,
+            fold: int | None = None,
+        ) -> float:
+            scoring_df = prepare_and_validate_scoring_df(
+                ground_truth, predicted, target, split, fold
+            )
+            gt_values = scoring_df[target].to_numpy()
+            pred_values = scoring_df[f"{target}_pred"].to_numpy()
+            return custom_calculation(gt_values, pred_values)
+
+    The metric function is automatically discovered and made available for
+    calculation. The function name after ``metric_`` becomes the metric name used
+    in the ``selected_metrics`` argument.
+"""
+
+import inspect
+import json
+import logging
+import sys
+from collections.abc import Callable
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+import polars as pl
+from scipy.stats import spearmanr
+
+from .assay import SEQUENCE
+from .dataset import Dataset, DatasetSlice, Subsets
+
+logger = logging.getLogger("proteingym.base")
+
+MetricFunction = Callable[..., float | None]
+
+_metric_functions_cache: dict[str, MetricFunction] | None = None
+
+
+def _split_slices(subsets: Subsets, split: str) -> list[DatasetSlice]:
+    """Return the list of slices for a named split strategy.
+
+    Args:
+        subsets: The Subsets object containing split information.
+        split: The name of the split strategy (e.g., 'random', 'kfold_random').
+
+    Returns:
+        The list of dataset slices associated with the split.
+
+    Raises:
+        TypeError: If the subsets slices are not stored as a dictionary.
+    """
+    if not isinstance(subsets.slices, dict):
+        raise TypeError(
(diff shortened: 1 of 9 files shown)


src/proteingym/base/evaluate.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Raises:
+        FileNotFoundError: If the prediction or dataset archive does not exist.
+        ValueError: If dataset_path is missing, if required parameters (--split,
+            --fold, --target) are missing for a Subsets file, or if --target is
+            missing for a plain Dataset.
+    """
+    logger.info("Start to calculate metrics.")
+
+    if dataset_path is None:
```
hredestig: why is it optional?

src/proteingym/base/evaluate.py
```diff
@@ -0,0 +1,127 @@
...
+    if dataset_path.name.endswith(".splits.pgdata"):
+        ground_truth: Dataset | Subsets = Subsets.from_path(dataset_path)
+    else:
+        ground_truth = Dataset.from_path(dataset_path)
+
+    predicted = Dataset.from_path(prediction_path)
+
+    test_fold: int | None = None
+    metrics_result: dict[str, Any]
+    if isinstance(ground_truth, Subsets):
```
hredestig: this if/else is a bit cumbersome to read in what needs to be given when - perhaps split to two functions, and make them two commands like `proteingym-base evaluate-splits`  and  `proteingym-base evaluate-data`

src/proteingym/base/evaluate.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+
+        metrics_result = {
+            "full_dataset": calculate_selected_metrics(
+                metrics_to_calculate, ground_truth, predicted, target, None, None
+            )
+        }
+
+    metadata = metrics_result.setdefault("metadata", {})
+    metadata["dataset"] = dataset_path.stem
```
hredestig: perhaps leverage pydantic here  `MetricsResult(model=model_name, split=split).model_dump_json(exclude_missing=True)`

src/proteingym/base/metrics.py
```diff
@@ -0,0 +1,464 @@
...
+
+
+def calculate_metrics_by_mode(
+    selected_metrics: list[str],
+    ground_truth: Subsets,
+    predicted: Dataset,
+    target: str,
+    split: str,
+    test_fold: int,
+    score_modes: list[str] | None = None,
```
hredestig: good case for `StrEnum`

src/proteingym/base/metrics.py
```diff
@@ -0,0 +1,464 @@
...
+) -> dict[str, Any]:
```
hredestig: how about returning a pydantic `BaseModel`?

src/proteingym/base/metrics.py
```diff
@@ -0,0 +1,464 @@
...
+
+
+def calculate_selected_metrics(
+    selected_metrics: list[str],
+    ground_truth: Subsets | Dataset,
+    predicted: Dataset,
+    target: str,
+    split: str | None = None,
+    fold: int | list[int] | None = None,
+) -> dict[str, float | None]:
```
hredestig, line 393: return `BaseModel`?

src/proteingym/base/metrics.py
```diff
@@ -0,0 +1,464 @@
...
+    ensuring complete prediction coverage. The returned DataFrame contains both
+    ground truth and predicted values for the specified target, aligned by
+    sequence and variables.
+
+    Args:
+        ground_truth: The ground truth data, either as a complete Dataset or
+            a Subsets object containing dataset slices.
+        predicted: The predicted Dataset containing model predictions for the
+            target. Must have the same structure (assays and variables) as the
+            ground truth.
```
hredestig: Could hint to `Dataset.predictions_delta` here as means to create that

tests/test_evaluate.py
```diff
@@ -0,0 +1,125 @@
+import json
+
+import pytest
+
+from proteingym.base.dataset import Dataset, Subsets
+from proteingym.base.evaluate import evaluate
+
+
+class TestEvaluateValidation:
```
hredestig: our best practice said we shouldn't use these classes

tests/test_evaluate.py
```diff
@@ -0,0 +1,125 @@
...
+            dataset_path=dataset_path,
+            selected_metrics=["spearman"],
+            model_name="test_model",
+            split="random",
+            target="DMS Score",
+            fold="0",
+            score_modes=["test"],
+        )
+
+        assert result_path == metric_path
```
hredestig: if pydantic, would be easy to turn this to a single `assert result == expection`

tests/test_metrics.py
```diff
@@ -0,0 +1,661 @@
...
+        corr = metric_spearman(
+            ground_truth=metrics_dataset_with_assay,
+            predicted=perfect_preds,
+            target="DMS Score",
+        )
+
+        assert corr == pytest.approx(1.0)
+
+
+class TestGetTopKFromSlice:
```
hredestig: test class

src/proteingym/base/evaluate.py
```diff
@@ -0,0 +1,167 @@
...
+        split: The evaluated split strategy, if applicable.
+        test_fold: The test fold index, if applicable.
+
+    Returns:
+        The path the metrics JSON was written to (same as ``metric_path``).
+    """
+    metadata: dict[str, object] = result.metadata or {}
+    metadata["dataset"] = dataset_stem
+    metadata["target"] = target
+    if model_name:
```
hredestig: you can skip  all these `if`, and the `json.dumps` piece - by just passing them, None or not, and then in `.model_dump(mode="json", exclude_none=True)`

karel-w: Can't do that if metadata is just a dict right? I've updated to a MetricsProvenance basemodel which can hold the metadata and DVC arguments we parse.

I also thought about extend MetricsResult but that has the downside if we want to make changes to our DVC method.

src/proteingym/base/metrics.py
```diff
@@ -0,0 +1,460 @@
...
+
+    Returns:
+        A MetricsResult with the computed modes populated and ``metadata`` describing
+        the test fold, the available train folds, and the total fold count. When
+        full_dataset mode is used, it scores against the complete dataset ignoring all
+        splits; the value is identical across folds since it evaluates the same data.
+    """
+    if not isinstance(context.ground_truth, Subsets):
+        raise TypeError("calculate_metrics_by_mode requires a Subsets ground truth.")
+    if context.split is None:
```
karel-w: What is a good way to deal with the following? @hredestig 

If I don't guard against context.split is None my typechecker complains because the ScoringContext allows for a none option (which is the case for evaluate_data), but since our evaluate_splits already requires a split this path is never reachable. But we have a few of these guard statements that are not tested (https://app.codecov.io/gh/ProteinGym/proteingym-base/pull/491) which just brings our score slightly below the 95%. 

Any ideas how to best rewrite?

hredestig: Could work around by `assert context.split is not None` (never expected to happen, but will evaluate. Not great practice but minor inconsistencies..) or just add marker to ignore the warning. But, if one wants to be tidier, then I think the problem is that we have this `ScoringContext` which has that complicated relationship between attributes. You could split to two classes instead, one that has `split` and one without and pass accordingly.

src/proteingym/base/metrics.py
```diff
@@ -0,0 +1,460 @@
...
+        raise ValueError("The scoring context must specify a split.")
+
+    if score_modes is None:
```
hredestig: actually here, you can also add this directly to the default instead like
```
score_modes = (ScoreMode.TEST, ScoreMode.TRAIN_AVAILABLE, ScoreMode.PER_FOLD)
```
(note the tuple instead of the forbidden mutable list) less if/else (so one test less) and easier to read in the docs. The type can then be `Sequence` instead to allow for passing a list if needed.

